PR #103, opened 2021-09-09
Bugfix: remoteControl CL argument not working

RebeccaNowak:
Fixed tiny bug that meant remote control app could not be started.


haraldsteinlechner:
seems fine for me

PR #254, opened 2022-11-03
cross platform openPath usage

AndrewAnnex:
* attempts to replace all usages of explorer.exe for electron equivalent

fsharp's syntax is fairly weird to me, so no doubt I am doing something incorrectly here either with the promise<string> returned by electron.

### DIFF
--- a/src/PRo3D.Core/SceneObjectsApp.fs
+++ b/src/PRo3D.Core/SceneObjectsApp.fs
@@ -6,6 +6,7 @@ open Aardvark.UI
 open Aardvark.VRVis
 open FSharp.Data.Adaptive
 open Adaptify.FSharp.Core
+open Aardvark.electron.shell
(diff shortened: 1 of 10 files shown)


haraldsteinlechner:
when looking at platform specific code i stumbled across those and fixed those without continuing this PR which i overlooked when trying digging deeper and deeper. I'm sorry for keeping you stuck here. As you might have noticed i'm trying to push more people (and me) to github so - hopefully this will not happen again...

The general problem behind the issue is that, the platform dependent code was run in the update function which has no electron at hand, since pro3d could be run with multiple views (looking at the same model).
Thus i moved all OpenFolder components into the UI part and refactored, which was quite an effort and goes further than your approach. https://github.com/pro3d-space/PRo3D/pull/267

AndrewAnnex:
no prob @haraldsteinlechner, closing in favor of #267

PR #265, opened 2022-12-12
Bugs/obj without texture

LauraFritz14:
obj updates in: wavefront loader, kdTrees, rendering


src/PRo3D.Viewer/Viewer/Viewer-Utils.fs
```diff
@@ -377,6 +379,7 @@ module ViewerUtils =
                     |> Sg.trafo trafo //(Transformations.fullTrafo surf refsys)
                     |> Sg.modifySamplerState DefaultSemantic.DiffuseColorTexture samplerDescription
                     |> Sg.uniform "selected"      (isSelected) // isSelected
+                    |> Sg.uniform "useTexcoords" (AVal.constant false) //(useTexcoords)
```
haraldsteinlechner: on aardvark style, uniforms should be written in capitals. I think this would be a large consolidation pass in Pro3D (also when we fix all wrong tabs)

haraldsteinlechner:
@chris-traxler if you are done, please merge, as you wish. Thanks Laura :)

chris-traxler:
Tested with 3 OBJs, 1) Dimorphos model with texture, 2) SLR test with vertex colours and 3) an arrow model without texture and vertex colours (see screenshots bellow). Import, annotations and picking worked with every imported OBJ. 
PRo3D_OBJ_1

PRo3D_OBJ_2

PRo3D_OBJ_3

Issue #252, opened 2022-10-27
screenshot folder selection does not work on macos


AndrewAnnex:
pressing the screenshot folder button in the UI causes the following error.
pro3d seems to assume explorer.exe exists for macos, which doesn't of course. 

```
 9: ERROR
 9:   [media] update function failed with: System.ComponentModel.Win32Exception (2): An error occurred trying to start process 'explorer.exe' with working directory '/Applications/PRo3D.Viewer.app/Contents/build/build'. No such file or directory
 9:      at System.Diagnostics.Process.ForkAndExecProcess(ProcessStartInfo startInfo, String resolvedFilename, String[] argv, String[] envp, String cwd, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
 9:      at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
 9:      at System.Diagnostics.Process.Start()
 9:      at System.Diagnostics.Process.Start(ProcessStartInfo startInfo)
 9:      at System.Diagnostics.Process.Start(String fileName, String arguments)
 9:      at PRo3D.SimulatedViews.ScreenshotAppModule.update(String baseUrl, Int32 numberOfSamples, String outputPath, ScreenshotModel m, ScreenshotAction action)
 9:      at PRo3D.ViewerApp.updateViewer$cont@1522-41(Model m, ScreenshotAction msg, Unit unitVar)
 9:      at PRo3D.ViewerApp.updateViewer(IRuntime runtime, IFramebufferSignature signature, BlockingCollection`1 sendQueue, FSharpMailboxProcessor`1 mailbox, Model m, ViewerAction msg)
 9:      at <StartupCode$Aardvark-UI>.$App.doit@72-2.Invoke(Unit unitVar0)
 ```

haraldsteinlechner:
thanks.

yes it is here: https://github.com/pro3d-space/PRo3D/blob/25e2de504b1870b88a41a3050fe65d5634524138/src/PRo3D.SimulatedViews/Screenshots/ScreenshotApp.fs#L73
it can be replaced by using electron's platform agnostic features - https://www.electronjs.org/de/docs/latest/api/shell to fix it.

PR #267, opened 2022-12-14
Bugs/explorer invocation

haraldsteinlechner:
fix for https://github.com/pro3d-space/PRo3D/issues/252

approach:
all invocations of explorer in update functions was replaced with electron commands in the view function. For surfaces this was quite complex since the scene path needed to passed all along to the ui components. rest was straightforward.

seealso: https://github.com/pro3d-space/PRo3D/blob/a3b8f3686a8878993b4e72e09076f323915ff3aa/src/PRo3D.Base/Utilities.fs#L1223
and https://github.com/pro3d-space/PRo3D/blob/a3b8f3686a8878993b4e72e09076f323915ff3aa/src/PRo3D.SimulatedViews/Screenshots/ScreenshotApp.fs#L83

This shoul be the way to go for all file interactions in the future. Also, when we need to support some of those operations in non-electron browsers (e.g. provex) this can easily be refactored to use replacements.

replaced explorer invocations for:
 * Scene Objects
 * Screenshots
 * Scene paths
 * Surfaces

@LauraFritz14 / @RebeccaNowak could you review the Scene Object related bits? 

Remaining occurrences are, not sure if those need to be ported.
 * remote control app
 * minverva

### DIFF
--- a/src/PRo3D.Core/Surface/SurfaceApp.fs
+++ b/src/PRo3D.Core/Surface/SurfaceApp.fs
@@ -34,7 +34,6 @@ type SurfaceAppAction =
 | MakeRelative              of Guid
 | RemoveSurface             of Guid*list<Index>
 | PickSurface               of SceneHit*string
-| OpenFolder                of Guid
 | RebuildKdTrees            of Guid
 | ToggleActiveFlag          of Guid
 | ChangeImportDirectory     of Guid*string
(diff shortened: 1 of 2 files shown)


haraldsteinlechner:
@RebeccaNowak could you please review this one @ThomasOrtner

RebeccaNowak (approved):

Issue #256, opened 2022-11-08
Adapt 'Locate surfaces' for obj files


sophie-sudoku:
**Problem**
At the moment .opc and .obj files are treated differently: 
Both can be referenced with absolute paths in the .pro3d file, but an invalid .opc file is ignored while an invalid .obj file throws an error and harms the scene-loading-process.  The 'locate surfaces' (action = ChangeImportDirectories) which updates the prefix of an import seems to be only applicable to .opc files.

**Solution**
It would be nice if .obj files are treated the same way as .obj files

haraldsteinlechner:
@LauraFritz14 what are your thoughts on this? i think you cleaned up at least the rendering and update code for objs vs opcs. Is this an easy one?

LauraFritz14:
pro3d version 4.10.3

Issue #246, opened 2022-10-21
Viewplans shall be stored in scene file


gpaar:
At present, any viewplans / rover positions that were either taken over from the trajectory or defined newly by "Place Rover" are not kept in the scene file for further use.
This shall be established.

LauraFritz14:
version 4.10.3

Issue #179, opened 2022-03-16
View Planner placed rovers not recorded in .pro3d and/or annotation file


gpaar:
Any placed rovers and other related settings are lost upon next PRo3D launch even if scene is saved.
As a "side error": The "show footprint" flag doesn't have any effect, same with the "export footprint" and "show footprint folder"

gpaar:
As a "side error": The "show footprint" flag doesn't have any effect, same with the "export footprint" and "show footprint folder"

LauraFritz14:
pro3d version 4.10.3

PR #271, opened 2022-12-18
viewplanner updates

LauraFritz14:
#256: reload obj to scene with wrong obj path
#179 and #246: viewplanner serialization


chris-traxler:
Successfully tested import of OBJ with wrong path and finding the right one (with SLR test scene).

View plan serialisation successfully tested for two view plans from traverse waypoints and two from rover positions.

PRo3D_View_Plans

PR #273, opened 2022-12-22
Replaces Adaptify.MSBuild with explicit adaptify invocation

haraldsteinlechner:
This PR changes the mode of operation for automatically generated model types. Instead of having the compilation integrated in the normal build process, here adaptify compilation needs to be triggered manually (see https://github.com/krauthaufen/Adaptify/blob/features/addGFilesToProject/README.md#local-mode).
The idea is to have better compile times, a more responsive IDE and better support for using vscode and other non-visual-studio tools.
The drawback is, that whenever a model type gets changed, a script need to be run to patch all automatically generated code.

At this stage, we are testing out the tradeoffs. I added @LauraFritz14  and @RebeccaNowak as reviewers for this PR - please test the branch in terms of development experience. 

- [x] compile everything using the adaptify tool
- [ ] ask for feedback
- [ ] change the docs accordingly


RebeccaNowak:
Tried out calling `dotnet adaptify --local --force --lenses --addToProject project.fsproj` to generate correct .g files, which worked. Not sure about compile time / responsiveness yet, need to test it some more. @haraldsteinlechner is it normal that the adaptify process still runs in the background?

haraldsteinlechner:
@RebeccaNowak actually i removed the adaptify msbuild references in the projects - https://github.com/pro3d-space/PRo3D/pull/273/commits/d74f6f3c5077564d8e08ae33c8f065b76487856a
this means - also the msbuild targets should have been removed => thus, no adaptify should be invoked. of course another old instance could have left a lurking adaptify in daemon mode (running in background, waiting for compile requests).
In the end we would commit all .g files, and the patched projects - only when changing model files, the tool needs to be invoked to update the .g files.

haraldsteinlechner:
is obsolete in terms of local adaptify approach

PR #279, opened 2023-01-11
Linux Support

haraldsteinlechner:
This one adds linux support. I could only test it on my machine at home (which i also used to compile native libs). I triggered the automatic deployment:

- [x] Testing the automatic deployment
pro3d-linux

It shall be noted, that currently linux users need to manually install libdevil-dev on their machine (see https://github.com/pro3d-space/PRo3D/tree/features/linuxSupport#beta-stage-linux-support)


haraldsteinlechner:
for robustness i needed to workaround: https://github.com/pro3d-space/PRo3D/issues/283 and upgrade all packages (rendering 5.3.0).

Issue #268, opened 2022-12-14
Sequenced Bookmarks blow up scene files (a lot a lot...)


haraldsteinlechner:
We need a practical solution to limit the size of the scene files when using bookmarks. Currently the bookmarks save all annotations into the scene repeatedly which results in enormous scene files.

This needs to be tackled asap. @ThomasOrtner  @RebeccaNowak  what are your ideas? 

In fact i think it also blew up a huge scene file when saving it which resulted in broken json:
![image](https://user-images.githubusercontent.com/513281/207662122-eb3bf549-6525-4cc5-8d44-6cb7a34159f4.png)

RebeccaNowak:
the requirement was to be able to change everything that can be adapted in the GUI via bookmarks
so I see two options: 
- we optimize the way things are saved (references, separate files, etc.) (time consuming)
- we drastically reduce what can be saved with bookmarks (e.g. only visibility) (simple but massively reduced functionality)
Or something in between those two.

haraldsteinlechner:
yes. the fundamental problem is that, when serializing the scene everything becomes a value. In memory, by using persistent data-structures we have a lot of sharing (for example each new annotation creates a new list of annotations, which refers to the old one).
In the serialization we loose this.. ,also, when loading the sharing cannot be restored - so this is also a runtime problem eventually.

This could be solved in general by doing the serialization smarter. This is a long-term thing. 
Clearly, we need a faster fix for people working with the feature.
That said, I think option (2) you mentioned seems to be a fruitful approach.

Why do you think it is that time consuming? couldn't we just generate a bookmark file per bookmark and load it lazily when activating the bookmark (in the model we only have the name of the bookmark file).

Pros:
 - i think this one is not that hard to implement
 - it is what you see is what you get

Cons:
 - we have a lot of files in a folder (one per bookmark)
 - it is not a general-purpose solution.
 - not sure about compatibility with old scenes, but should be doable via versioning?

What do you think?

RebeccaNowak:
Only generating a bookmark file per bookmark doesn't sound too time-consuming. Yes, I think compatibility should be ok with versioning.

haraldsteinlechner:
i think, this is the spot:
https://github.com/pro3d-space/PRo3D/blob/25e2de504b1870b88a41a3050fe65d5634524138/src/PRo3D.Viewer/Viewer/Viewer-IO.fs#L260 to:

- serialize bookmarks separately
- set a the file name in the scene

[...]

PR #282, opened 2023-01-17
Bugs/bookmark size

RebeccaNowak:
Bookmarks are now saved in separate files, to avoid creating overly large scene files. Bookmarks are saved in a folder with the same name as the scene file. Loading old scenes should work without issue (and is tested for scene files created with the current develop version), but if someone has an old scene file lying around it would be great if you could test it ( @LauraFritz14 @haraldsteinlechner )


haraldsteinlechner:
btw. this is the fix for this https://github.com/pro3d-space/PRo3D/issues/268, right?

RebeccaNowak:
Yes, this is the fix for Issue #268

Issue #167, opened 2022-02-17
Radiometry Support


chris-traxler:
Support radiometry in the following way for an individual surface:
1st solution allow to insert 6 values ar, br, ag, bg, ab, bb to linearly modify the color mapping of rgb individually
![image](https://user-images.githubusercontent.com/65888251/221801830-5a48b399-8f86-45ee-88e3-9d07959a9cbb.png)
Default values shall be memorized as per scene
Optionally / later
- Export / import LUTs as CSV (RGB separate channels)
- Function to automatically stretch histograms per channel (to remove the reddish hue)
- Take over values from meta data as soon as they exist (average / sigma stretch mode)

haraldsteinlechner:
Basic functionality is approved, optional features for later:
Optionally / later
- Export / import LUTs as CSV (RGB separate channels)
- Function to automatically stretch histograms per channel (to remove the reddish hue)
- Take over values from meta data as soon as they exist (average / sigma stretch mode)

PR #339, opened 2023-06-27
#167 radiometry: shader update

LauraFritz14:
shader update


LauraFritz14:
obsolete, fix in Rebeccas pr

Issue #308, opened 2023-02-28
Distance Filter


gpaar:
Limit distance from OPC origin (Pivot point of surface once known)

LauraFritz14:
Mail von Gerhard 20.4.2022:
Es ist hier die Distanz zur Aufnahmeposition des betreffenden OPCs gemeint. D.h. der Distanzfilter sollte Surface-spezifisch angebracht werden (ähnlich wie der Triangle-Filter, oder die Transformation).
Leider haben wir aktuell noch keine Metadaten-Struktur, aber man könnte den Pivot-Punkt der Surface verwenden, wie er zurzeit definiert wird.
Omit all quads of the respective surface from display that contain at least one vertex that exceed the specified distance from the pivot point. Do not take into account those quads for measurements / annotations either.

ThomasOrtner:
Why not use the home position? The pivot point can only be picked on the surface so it would be impossible to filter for instance from the Mastcam position as given by the traverse. Also there is an abandoned branch where I started this (https://github.com/pro3d-space/PRo3D/tree/features/130_triangleDistanceFilter) but sth. was wrong in the shader and I ran out of time ;)

gpaar:
That's the plan. We're about to generate some meta data that should also contain that home position (= the center of our spherical projection that forms the OPC). Pivot point will be the first workaround.

ThomasOrtner:
At the moment, every surface gets the center of its bounding box as default "home position". The user can then set the current camera position and look at as new "home position" via the `Surface Properties` by clicking the button next to "Set Home Position". This "home position" is also the position and look at that the camera transitions to when the house icon is clicked to trigger a goto animation.

PR #359, opened 2023-09-20
Feature/308 distance filter

LauraFritz14:
#308 Distance Filter
- select opc
- in surface properties do:
- set home position (not working without)
- check DistanceFilter
- set FilterDistance

### DIFF
--- a/src/PRo3D.Viewer/Viewer/Viewer-Utils.fs
+++ b/src/PRo3D.Viewer/Viewer/Viewer-Utils.fs
@@ -385,6 +385,29 @@ module ViewerUtils =
                     )
                     |> Sg.dynamic
 
+                let homePosition =
+                    adaptive {
+                        let! homePosition = surf.homePosition
+                        
+                        match homePosition with
+                        | Some hp -> 
+                            return hp.Location
+                        | None ->
+                            let! bb = surface.globalBB
+                            return bb.Center                        
+                    }               
+                    
+                let filterByDistance =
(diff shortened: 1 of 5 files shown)


src/PRo3D.Viewer/Viewer/Viewer-Utils.fs
```diff
@@ -701,23 +728,41 @@ module ViewerUtils =
...
-                    yield input.P2
+                let filterDistanceActive : bool = uniform?FilterByDistance
+                let triangleSizeCheck = (alpha && beta && gamma)
+
+                if filterDistanceActive then
+                    let filterRange : float = uniform?FilterDistance
+                    let homePosition : V3d = uniform?HomePosition
+
+                    let inRange = 
+                        (Vec.Distance(homePosition, input.P0.wp.XYZ)) < filterRange &&
```
haraldsteinlechner, line 739: is wp really world position? this is numerically not stable?

haraldsteinlechner:
as hinted above, should filtering operate on view space as opposed to wp which is used here? i think there is a numerical issue. if so i'd propose:
 - transform homePos to view space, rename it to homePosViewSpace
 - compare it with vp instead of wp

LauraFritz14:
Thank you Harri! I updated the branch with your hints

haraldsteinlechner:
Excellent! I'm merging this one. If there is an asscoiated issue please close it.

PR #362, opened 2023-10-17
Features/296 trafo

LauraFritz14:
trafo with pivot for each surface. keeps translation and resets scaling and rotation if pivot changes.
how to use:
- select a surface
- enable "use pivot" and "show pivot" in transformation menu for surface
- in picking actions set "PickPivot" for picking a pivot position with Strg + LMB
- adjust transformation via transformation gui

### DIFF
--- a/src/PRo3D.Core/Transformation-Model.fs
+++ b/src/PRo3D.Core/Transformation-Model.fs
@@ -40,6 +40,7 @@ type Transformations = {
     scaling               : NumericInput
     trafoChanged          : bool
     usePivot              : bool
+    pivotSize             : NumericInput
 } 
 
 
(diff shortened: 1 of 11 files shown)


src/PRo3D.Core/Transformation-Model.fs
```diff
@@ -285,6 +300,7 @@ module Transformations =
                 scaling              = scaling
                 trafoChanged         = false
                 usePivot             = false
+                pivotSize            = Initial.initPivotSize 4.0
```
haraldsteinlechner, line 303: it might be a good idea to factor out readPivotSize which is then used top level where version dispatching occurs. What do you think?

haraldsteinlechner:
looks good. I'd suggest writing a little docu about pivot similar to https://github.com/pro3d-space/PRo3D/blob/develop/docs/Feature-Multitexture.md since there is more complex stuff coming up with this feature...

PR #368, opened 2023-11-29
Features/opc tool

haraldsteinlechner:
extended opc-tool with dds functionality and dotnet-tool packaging, see: https://github.com/pro3d-space/PRo3D/blob/features/opc-tool/docs/OpcTool.md

touched files: build scripts, and src/opc-tool.


src/PRo3D.Core/Surface/Surface.Files.fs
RebeccaNowak, line 1: Should the variable name be quiet instead of quite? Since it controls logging output?

` let isOpcFolder' (quite : bool) (path : string) = `

RebeccaNowak (commented):
found one possibly confusing argument name "quite" instead of "quiet" (added note in review)

src/PRo3D.Core/Surface/Surface.Files.fs
haraldsteinlechner: fixed in: https://github.com/pro3d-space/PRo3D/pull/368/commits/6095bf0427bd10e7f3271b36834f2cc9b3c6f1e3

Issue #358, opened 2023-09-12
Bug: Batch processing / Sequenced Bookmarks Frustum


RebeccaNowak:
With the same settings in GUI and Batch Rendering, the field of view of rendered images is not the same as shown in the GUI.
Focal length is the same, calculated Frustum is not.

Depending on the aspect ratio of the output image the field of view is wider or narrower in the rendered images than in the GUI.

We have one Frustum directly in Model, and another one in ViewConfigModel. Those two have different values for FoV. The reason for this is, that the frustum in ViewConfigModel is not updated when the size of the RenderControl is changed. Snapshots use the value from ViewConfigModel, but in Viewer the one in Model is used.

RebeccaNowak:
The underlying problem was this:

```

Frustum.perspective (hfov.DegreesFromRadians()) near far 1.0
        |> Frustum.withAspect aspect
```

and 

`        Frustum.perspective (hfov.DegreesFromRadians()) near far aspect`

do not have the same result, but were being used interchangeably.

Demonstration of this outside PRo3D:
 ```
3: WARNING: Frustum.perspective 60.0 0.1 100.0 0.8
 3: Frustum Model: { left = -0.05773502692
 3:   right = 0.05773502692
 3:   bottom = -0.07216878365
 3:   top = 0.07216878365
 3:   near = 0.1
 3:   far = 100.0
 3:   isOrtho = false }
 3: Frustum Model Aspect: 0.8
 3: Frustum Model Fov: 59.99999999999999
 3: WARNING
 3:   Frustum.perspective 60.0 0.1 100.0 1.0
 3:                            |> Frustum.withAspect 0.8
 3: Frustum Model: { left = -0.04618802154
 3:   right = 0.04618802154
 3:   bottom = -0.05773502692
 3:   top = 0.05773502692
 3:   near = 0.1
 3:   far = 100.0
 3:   isOrtho = false }
 3: Frustum Model Aspect: 0.8
 3: Frustum Model Fov: 49.58256179428978
```

PR #370, opened 2023-12-12
Bugs/snapshot frustum  #358

RebeccaNowak:
Bugfix #358
Frustum.withAspect from Aardvark.Media has unexpected behaviour. Replaced calls to the function.
Now updating Frustum in FrustumModel to be consistent with Model.frustum.
Also setting formerly unused Model.aspect on resize of rendering control.

### DIFF
--- a/src/PRo3D.Base/Utilities.fs
+++ b/src/PRo3D.Base/Utilities.fs
@@ -1329,6 +1329,13 @@ module Electron =
         sprintf "top.aardvark.electron.shell.showItemInFolder('%s');" (JsInterop.escapePath s)
 
 module FrustumUtils =
+    let withAspect (aspect : float) (frustum : Frustum) =
+        Frustum.perspective 
+            (Frustum.horizontalFieldOfViewInDegrees frustum)
+            frustum.near
+            frustum.far
+            aspect
+
     let calculateFrustum (focal : float) (near : float) (far: float) =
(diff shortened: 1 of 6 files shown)


haraldsteinlechner:
for the record, what is the reason for storing aspect in the model?

RebeccaNowak:
I suspect that is has historical reasons, that we save the frustum in Model AND FrustumModel. Both are used in the code in different places, but I think we could remove Frustum from Model and only store it in FrustumModel. I could open a new Issue with the task to look into this and clean it up, if you don't see a problem with removing Frustum from Model.

haraldsteinlechner:
Hm, cool thanks for explaining. Taking it further, conceptually, the model should be "viewer" independent (so when e.g. a browser also opens a connection there is no race for updating the frustum). So best would be to have no width/height/aspect dependent stuff in the model at all. Since this seems unrealistic (is it?), your approach in removing redundancy would be a good option. I'm not sure if it pays out though the effort (given the fact that having aspect dependent stuff in the model is not preferred anyways).

So let's go with this PR and let's create an issue which also talks about the possibility to really remove the aspect as a whole as a long-term objective.

What do you think?

RebeccaNowak:
Sounds good to me! Not sure yet what the implications of completely removing the aspect would be, but we could definitely look into this.

Issue #344, opened 2023-07-04
Persistent (Hanging) panning in 3D view when leaving the window


chris-traxler:
When panning with the middle mouse button in the 3D view and releasing it outside of the 3D view (e.g. in the UI), the release is not recognised. When returning to the 3D view, panning continues without pressed middle mouse button.

Workaround: Click the middle mouse button in the 3D view to stop panning.

haraldsteinlechner:
Related, partly subsumed: https://github.com/pro3d-space/PRo3D/issues/192 and https://github.com/pro3d-space/PRo3D/issues/137

My proposal is to upgrade to the new orbit controller - to get rid of deprecation, smoother animation, less threads and maintained pointer apture solution.

The pointer capture problem in free fly is still unclear though.

RebeccaNowak:
Should be fixed with pull request #423

RebeccaNowak:
More background: FreeFlyController does not seem to recognize when the middle mouse button is released if the cursor is outside the rendering view AND while the middle mouse button is pressed. It works with right and left mouse button. The problem is not PRo3D-specific, it can also be observed in Aardvark.Media in the GoldenLayout Demo. The problem does not occur in the NavigationDemoNew although both use FreeFlyController.

PR #423, opened 2024-10-07
fix issue with panning #344

RebeccaNowak:
now manually setting pan to false on EndDragging (also works for middle mouse button / pan)


RebeccaNowak:
What's going on in a broader sense: The middle mouse button is used by browsers for various actions, and they sometimes (and depending on the browser of course) interfere when you want to use middle mouse button events. Also, browsers sometimes even override/ignore "preventDefault()" for these actions. I suspect that's why the mouseOut event does not fire while the middle mouse button is pressed (I tried this out). So we can't use mouseOut / mouseIn to solve this. However, onMouseUp does fire and so we can use this.

As for the multiselectBox: I didn't know anything about that, so I left the functionality as it was. If it's not in use any more we can remove that code entirely.

haraldsteinlechner:
Let's merge this one - but let's keep the issue open, after updating the media packages we can clean up the code and remove the unnecessary handlers.

PR #428, opened 2024-10-16
Features/426 load more than one traverse

ThomasOrtner:
- enabled traverse import dialog to select multiple files
- traverse can now be added to existing traverses
- removed default text and dots to reduce loading times
- added delete all action button to clear traverses
- added natural sorting because of route numbering
- made traverse name and guid non-adaptive

### DIFF
--- a/src/PRo3D.Viewer/TraverseApp.fs
+++ b/src/PRo3D.Viewer/TraverseApp.fs
@@ -2,7 +2,6 @@
 
 open System
 open System.IO
-open System.Text.RegularExpressions
 open Aardvark.Base
 open Aardvark.UI
 open Aardvark.UI.Primitives
@@ -12,9 +11,6 @@ open PRo3D.Base
 open PRo3D.Core
 open FSharp.Data.Adaptive
 open FSharp.Data.Adaptive.Operators
-
-
-
         
 module Double =
     let parse x =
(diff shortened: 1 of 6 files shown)


src/PRo3D.Viewer/TraverseApp.fs
```diff
@@ -347,34 +338,64 @@ module TraverseApp =
...
+        let s1 = left.tName|> AVal.force
+        let s2 = right.tName|> AVal.force
+
+        let rec compareParts (parts1: string list) (parts2: string list) =
+            match parts1, parts2 with
+            | [], [] -> 0
+            | [], _ -> -1
+            | _, [] -> 1
+            | h1::t1, h2::t2 ->
+                let isNum1 = Int32.TryParse(h1)
```
haraldsteinlechner, line 359: bit nitpicking but thinking about this one...

would creating a list of char and using default comparison work also?
```
> [1;2;3] < [2;3;4];;
val it: unit = ()

> val it: bool = true

```

also, how about having the tool general purpose, e.g. of string and provide it with `(adaptiveTraverse1.Current, adaptiveTraverse2.Current) ||> AVal.map2 (fun l r -> naturalSort l.tName r.tName)`?

src/PRo3D.Viewer/TraverseApp.fs
```diff
@@ -347,34 +338,64 @@ module TraverseApp =
...
-                let traverse = Traverse.initial name sols 
-                let traverses' = HashMap.add traverse.guid traverse model.traverses
-
-                { model with traverses = traverses'; selectedTraverse = Some traverse.guid }
-            else
-                model
+        | LoadTraverses paths ->            
+            let traverses =             
+                paths 
+                |> List.filter(File.Exists)
```
haraldsteinlechner, line 380: having warning for ignored paths would be nice

haraldsteinlechner (commented):
cool thanks. added some minor points for potential improvement

ThomasOrtner:
- made natural sorting comparer to work with strings and moved it to PRo3D.Base Utilities.fs
- added a warning if traversal files are not existing

PR #479, opened 2024-12-09
Features/440 feature queries produce unexpected obj cutouts

ThomasOrtner:
Investigated Notebook based Annotation Queries for cutting out geometry from OPCs to explore ICP co registration. The query API has a flaw and a bug:
- bug: ROI patch filter does not return all patches inside the ROI
- flaw: cutouts spanning multiple patches are returned in a multi object obj (MOO), basically no standard obj viewer can correctly read these files although MOOs are official obj standard. 

This PR neither fixes the bug nor addresses the flaw. I deactivated the patch filter to return all patches of all patchhierarchies, which is potentially slow but correct. As I only need the points for ICP I added a raw point cloud mode to the payload json returning all points inside the ROI in a single object obj. 

I understand that these feature queries are an experimental feature, so also this issue is rather an investigation than an actual enhancement or fix 😄 The patch filter still needs to be fixed


haraldsteinlechner:
this PR was based on main and diverged heavily from develop (due to package version upgrades). i rebased all those changes onto develop in a separate branch and PR, this this one is a duplicate.

PR #480, opened 2024-12-10
Features/440 feature queries produce unexpected obj cutouts rebased

haraldsteinlechner:
I rebased all changes of https://github.com/pro3d-space/PRo3D/pull/479 to develop. The problem with the original PR is that it is based on main, and modified a lot of lines which have been affected in develop by switching to new base packages. This made rebasing necessary. Unfortunately including manual work fixing the conflicts. I will go through the condensed changes here in detail.

@ThomasOrtner 's original PR:

"Investigated Notebook based Annotation Queries for cutting out geometry from OPCs to explore ICP co registration. The query API has a flaw and a bug:

- bug: ROI patch filter does not return all patches inside the ROI
- flaw: cutouts spanning multiple patches are returned in a multi object obj (MOO), basically no standard obj viewer can correctly read these files although MOOs are official obj standard.

This PR neither fixes the bug nor addresses the flaw. I deactivated the patch filter to return all patches of all patchhierarchies, which is potentially slow but correct. As I only need the points for ICP I added a raw point cloud mode to the payload json returning all points inside the ROI in a single object obj.

I understand that these feature queries are an experimental feature, so also this issue is rather an investigation than an actual enhancement or fix 😄 The patch filter still needs to be fixed"


haraldsteinlechner:
https://github.com/pro3d-space/PRo3D/pull/480/files#diff-8087808cea81830bd69d9d4f1e69d76c9ed2287ebfda11344b07d363d7a192f7R316 this one leads to double loading of patch hierarchies, but getting rid of this is part of another rework and will be subsumed by faster-reordering of surface priorities track of work. So i'm gonna merge this one to unblock the other efforts.

Regarding the points of the original PR, i batched together all objects into a single obj (which has it drawbacks also such as textures are harder), and switched the culling method to bounding box based. For my test cases this works (unfortunately the original code also works so i cannot reproduce your issue and why you started investigating in the first place.

PR #483, opened 2024-12-17
Reimplemented lod-metric

haraldsteinlechner:
There are still two major problems, which cannot be tackled without reworking larger parts of the rendering code:
- for large zoom's many patches get considered due to their poor AABB and cannot be sorted out without more metadata or data-guided LoD.
- the new lod metric uses triangle size as estimate for geometric complexity. oddly sized huge triangles in patches increace the average triangle size leading to large memory footprint without providing additional detail in areas of large triangles. as a workaround i use the old lod metric also to at least not pull in more data as before the PR (i'm not happy with it but there is no time reworking really large parts and re-computing meta-data)

### DIFF
--- a/Build.fs
+++ b/Build.fs
@@ -237,7 +237,7 @@ let test = """let viewerVersion       = "3.1.3" """
     res
     *)
 
-let aardiumVersion = "2.0.5"
+let aardiumVersion = "2.1.1"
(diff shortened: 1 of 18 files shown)


haraldsteinlechner:
was included in this one.

Issue #412, opened 2024-08-20
Depth Map Batch Generation


gpaar:
Provide a PRo3D plug-in to obtain reference depth maps from given viewpoints, the bounding box and geometry of a spherical and/or cylindric (both in azimuth / elevation ROI) and/or perspective view (given by a quaternion) coordinate system mosaic shall be a relevant input parameter

RebeccaNowak:
Additional information from Meeting: 
Depth map panos should be created
Input: 
Sphere or cylinder with bbox to define mapping -> cylinder related to scalebars and Mars 2020, sphere for angular position, medium prio 
A set of perspective views (quaternions) -> higher prio, stitching of single images

chris-traxler:
Actually this has changed in the meeting today. Spherical projection has the highest priority. Beside a depth map pano also a pano with the image texture should be created using spherical projection.

RebeccaNowak:
@gpaar 
For context, could you describe your use case a little? What kind of data do you derive your input from, and what do you need the panoramic depth maps for?

We understand the input to PRo3D will be a bounding box and geometry of a spherical view (in azimuth / elevation ROI).
This gives us three options:
1. input one set of parameters in the GUI, and create one set of results (one depth map, one RGB texture / normal rendering)
2. input a json file with multiple spherical views via GUI, and create a depth map & normal rendering for each of them in batch rendering mode (running in the background as with sequenced bookmarks)
3. the same as (2) only via command line, without using the GUI 

If option (2) or (3): Do you already have a format you use to export spherical views? If not, do you have preferences what the format we use to exchange them should look like?

Should this feature be in any way connected to bookmarks, or should it be entirely independant? For example, use bookmarks to define spherical views, or create bookmarks at imported views.

Colour of the depth maps: Should it be red to blue, or do you need greyscale or a different colour scheme?

gpaar:
No connection to bookmarks required.
However, spherical & cylindric export from bookmarks (as well as spherical & cylindric display in main display widget) could be considered in future, not as content of this thread.

json file (option 2) is preferred with format to be defined by implementing entity.

Export format: float Tif with actual distance values coded in the pixel values, plus an exported json that defines the center and extent, with preference on JR-GPC. Example attached, the ModelType DemSphere is the one to use. Use AttributeLayers and Texture if texture is being exported as well, same with other future attributes to be optionally exported.
[...]

PR #496, opened 2025-02-19
Features/depth panoramas #412

LauraFritz14:
depth panoramas generation:
- start viewer, load a scene and set sequenced bookmarks
- start the depth panoramas generation via the "Generate Depth Images" button in the "Depth Panoramas" gui
- a panoramaInputFormat.json file is generated in the folder specified in the "Output Path"
- the snapshot.exe starts and generates the panorama depth images for each bookmark in the list

output:
3 images for every direction of each bookmark
- *.png = color image
- *.png_depth = scaled depth download (between 0 and 1)
- *.tiff = real distance (in negative z-direction)


haraldsteinlechner:
@LauraFritz14 i remember you mentioned there are some extensions planned (json files), is this already included here and finished?

PR #500, opened 2025-03-03
Features/489 endpoints for notebook based co registration

ThomasOrtner:
ah sorry, wrong base branch, now with merge in develop.


haraldsteinlechner:
we merged major trafo rework PR into develop, leading to those conflicts. Since the trafo stuff is quite criticial, i would suggest you merge develop and test your functionality again with those changes.

src/PRo3D.Viewer/Properties/launchSettings.json
```diff
@@ -7,7 +7,7 @@
     },
     "PRo3D.Viewer.Provex": {
       "commandName": "Project",
-      "commandLineArgs": "--disableCors --port 4321 --remoteApi -loadRecent",
+      "commandLineArgs": "--disableCors --port 4321 --remoteApi --scene \"F:\\PRo3D\\AI-Mars-3D\\Co-Registration\\2025-01-07_CoRegDemo\\2025-01-07_Kodiak_CoReg_Demo_SRC_translated.pro3d",
```
haraldsteinlechner, line 10: when using hardcoded paths, please use a specific launch setting

src/PRo3D.Viewer/RemoteApi.fs
```diff
@@ -759,6 +811,57 @@ module RemoteApi =
                         path "/queryAnnotationAsObj" >=> request (queryAnnotationAsObj api)
                     ]
                 )
+                prefix "/exp" >=> 
+                    choose [
+                        path "/greet" >=>
```
haraldsteinlechner, line 816: is this just testing code or a real endpoint

haraldsteinlechner (changes requested):
thanks for the great pr and all the cleanups, i just noted some minor things which would be nice to be fixed.

src/PRo3D.Viewer/Properties/launchSettings.json
haraldsteinlechner: it is not the best way either, but i started to do specific tests like this, leaving core startup launch settings clean: 
image

src/PRo3D.Viewer/RemoteApi.fs
ThomasOrtner: testing code

src/PRo3D.Core/Queries/AnnotationQuery.fs
```diff
@@ -324,7 +323,7 @@ module AnnotationQuery =
                 {
                     colors = None
                     indices = [| |] // h.indices |> Seq.toArray
-                    vertices = allVertices |> shiftByCenterOfMass
+                    vertices = allVertices //|> shiftByCenterOfMass
```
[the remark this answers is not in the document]
ThomasOrtner: kind of yes, paraview and some other tools really dont like large coords. this should be an option though instead of using comments :D

Issue #492, opened 2025-01-29
Distance to surface point


chris-traxler:
In the simulated view of the View Planner it should be possible to get the distance from the current viewpoint to a selected surface point. This can be based on single point measurements. The distance should be shown beside the picked point in the simulated view.

Issue #467, opened 2024-11-27
Viewplan parameters into Scene


gpaar:
Enable storing and recalling viewplan parameters in the scene (flags and parameters).
Each viewplan should also store its current pan/tilt settings. In case fals color settings are possible to store as per viewplan position, these should be kept resident likewise.

Issue #404, opened 2024-08-13
Long-Baseline View Planning Upgrades


chris-traxler:
The View Planner should be extended to support instruments of the Curiosity rover.

**Tasks**

- [x] Support refined MSL Mastcam camera models --> trivial / already solved
- [x] Include MSL MAHLI camera model --> Modified to ChemCam RMI: JR to provide camera model once ready
- [x] Allow target-based selection of pan-tilt values. This means that these camera parameters are associated with targets and when selecting one on the surface, the camera’s orientation is set automatically.

PR #504, opened 2025-03-31
Features/viewplan updates

LauraFritz14:
#467: footprint added to each viewplan

#492:  you can add distance points to selected vp
- select "PickDistancePoint" in picking interactions
- CTRL+LM to place point on surface
- there is a "DistancePoints" part in the vp menu where the points are listed and some self-explanatory settings

#404: if you click on the house icon beside a distance point in the list, the rover turns his lookAt direction to this point (approach to simulate a target point)


src/PRo3D.Core/Model.fs
```diff
@@ -23,6 +23,7 @@ type Interactions =
     | PlaceSceneObject      = 17
     | PickPivotPoint        = 18
     | PickSurfaceRefSys     = 19
+    | PickDistanePoint      = 19
```
haraldsteinlechner, line 26: two items with the same tag

PR #507, opened 2025-04-07
Features/modelviewer v2

sophie-sudoku:
i updated the models in the submodule - i changed the references here accordingly

### DIFF
--- a/src/ModelViewer/Properties/launchSettings.json
+++ b/src/ModelViewer/Properties/launchSettings.json
@@ -6,15 +6,23 @@
     },
     "AFC Camera": {
       "commandName": "Project",
-      "commandLineArgs": "..//..//..//resources//Hera-Spacecraft//gltf//AFC ..//..//..//resources//Cubemap-Mars"
+      "commandLineArgs": "..//..//..//resources//Hera-Spacecraft//Hera//gltf//AFC ..//..//..//resources//Cubemap-Mars"
     },
     "Hera Spacecraft": {
       "commandName": "Project",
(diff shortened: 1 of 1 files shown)


haraldsteinlechner:
you updated the submodule but i don't see a changed submodule ref in the commit. did you forget to commit the changed ref?

ThomasOrtner:
Please review what needs to be done here to close this PR. File changes only contain changed launch settings for probably loading different models

sophie-sudoku:
sorry, i oversaw the comments!

the reference did not change

[submodule "src/ModelViewer/resources"]
	path = src/ModelViewer/resources
	url = https://github.com/pro3d-space/PRo3D.Resources.Models

only the content. so i updated the paths that are referenced in the profiles. to update the submodule content run 

git submodule update --remote

i just tested it again, the models are correctly loaded

sophie-sudoku:
i wasnt aware that setting a new submodule reference (updating to a different commit) has to be committed as well - done :)

sophie-sudoku:
i do not understand why the checks fail after updateing the submodule reference, as only the reference was updated and nothing else. The fail is within the scope of **Unchanged files with check annotations** as i see it... i tried rerunning the checks, which did not resolve the problem

haraldsteinlechner:
the checks fail because the branch is based on an outdated version of aardvark.fontprovider, which results in [compile errors](https://github.com/pro3d-space/PRo3D/actions/runs/17042895493/job/48310969612?pr=507#step:4:675). This was fixed in develop branch already by upgraded the dependency -> this is why the PR checks succeed.

haraldsteinlechner (approved):
the build is fixed in the resulting merge, so this is just the submodule update and the startup arguments.

PR #521, opened 2025-06-04
Features/404 target based pan tilt

LauraFritz14:
PanTiltTarget

How to use the targets:
1. Place a ViewPlan: Set the PickingInteractions to "PlaceRpver" and click (1) for the rover position and (2) the lookAr direction. 
2. Select the ViewPlan in the listing to adjust the properties
3. Select a camera to see the InstrumentView
4. Set the PickingInteractions to"PickDistancePoint" and Ctrl + click to set DistancePoints. Click on the house button next to the distance point, named by its distance, to set the rover's pan and tilt.


ThomasOrtner (changes requested):
Functionality is nice, but it appears that
* the text size is not updated once a point is placed
* it is quite a good idea to use the distance as an identifier for now, but i would still limit it to 4 digits after the comma

otherwise this is fine for, please implement these changes before approving https://github.com/pro3d-space/PRo3D/pull/523

LauraFritz14:
Thank you for the review, all suggested changes have been implemented.

PR #523, opened 2025-06-12
Features/518 surface trafo export import

LauraFritz14:
To export a surface trafo, select the surface, adapt the trafo and click on the "export TrafoData" button. The json file with the trafo data is stored in the respective surface folder.

importTrafo

To import the surface data select the surface and use "Extras -> Import Transformation (*.json)" in the main menu.


ThomasOrtner:
functionality appears to work fine, format is also fine. Interoperability  with other tools is a bit difficult with pivot point and east north intricacies. We will see if there will be usecases like this. 

* I would still suggest to put import and export into the properties making an row label Trafo I/O and making an export and an import button. It is a bit unnatural to first select a surface and then go to extras. 
* further I would also do a file dialog for the export opening at the surface folder, also the import dialog should open at the surface folder

src/PRo3D.Core/ScaleBarsApp.fs
```diff
@@ -315,7 +315,7 @@ module ScaleBarsApp =
                 let scB = model.scaleBars |> HashMap.tryFind id
                 match scB with
                 | Some sb ->
-                    let transformation' = (TransformationApp.update sb.transformation msg refSys)
+                    let transformation' = (TransformationApp.update sb.transformation "" msg refSys)
```
ThomasOrtner: better use an option instead of empty string

src/PRo3D.SimulatedViews/Viewplanner/ViewPlanApp.fs
```diff
@@ -884,18 +944,47 @@ module ViewPlanApp =
             match model.selectedViewPlan with
             | Some vpid -> 
                 let selectedVp = model.viewPlans |> HashMap.find vpid
-                let selPoint = selectedVp.distancePoints |> HashMap.find id
-                
-                let footPrint = Optic.get _footprint outerModel
-                let forward, right, up = decomposeRoverTrafo selectedVp.roverTrafo
-                let forward = (selPoint.position - selectedVp.position).Normalized
+                //let selPoint = selectedVp.distancePoints |> HashMap.find id
```
ThomasOrtner: better remove instead of commented

src/PRo3D.Core/TransformationApp.fs
```diff
@@ -256,8 +311,9 @@ module TransformationApp =
 
     let update<'a> 
         (model : Transformations)
+        (exportPath : string)
```
ThomasOrtner: this feels like it should be outside of the transformationapp and not a parameter, but I understand this might be complicated, lets leave it like that ... suggestions @haraldsteinlechner ?

src/PRo3D.Viewer/Viewer/Viewer.fs
```diff
@@ -1001,6 +1001,19 @@ module ViewerApp =
         | ImportTraverse traverseFiles,_,_ -> 
             let t = TraverseApp.update m.scene.traverses (TraverseAction.LoadTraverses traverseFiles)
             { m with scene = { m.scene with traverses = t }}
+        | ImportTrafo trafoFiles,_,_ ->
```
ThomasOrtner: we can match directly here to a single list element
```
| ImportTrafo([trafoFile], _, _) -> 
    // Only matches if there is exactly one trafo file
    ...

```

ThomasOrtner (changes requested):
This PR also seems to mix in changes and updates from the camera pointing feature .. https://github.com/pro3d-space/PRo3D/issues/404 target based pan tilt update ... not a problem, but we should just make sure to check https://github.com/pro3d-space/PRo3D/pull/521, before approving this PR

suggested changes:
* make export path an option
* move export from the menu and make a dialog for the export within surface properties... * two buttons, preferably with icons

LauraFritz14:
> * I would still suggest to put import and export into the properties making an row label Trafo I/O and making an export and an import button. It is a bit unnatural to first select a surface and then go to extras.

Yes, I would also find it better that way, but it is technically not possible at the moment. We already had this problem when loading images for the projected images in the view plans. Somehow the action gets lost. Harry also tried it back then and came to the same conclusion as far as I remember. We thought it might be a problem with the tabs...

LauraFritz14:
Update: chooseFile button added to Transformation tab (and deleted from "Extras" in main menu)

PR #543, opened 2025-10-14
Support for additional bodies (MOON, PHOBOS, DEIMOS)

haraldsteinlechner:
This is a simple one and extends the planet/body dropdown with predefined bodies. Note that this is not really flexible yet, for multiple bodies PRo3D-GIS version needs to be used.


haraldsteinlechner:
this one also fixes broken deployment due to yarn updates..

PR #556, opened 2025-11-03
Bugs/hera snapshot config

Brunnhuber:
There are some Problems with the creation of Images created with Pro3D Snapshots. Some Configurations are not used correctly and the image quality seems to be worse than before.

Testing of the different arguments and fixing where they show unexpected behavior.


Brunnhuber:
Fixing configuration of Hera Snapshots. 

- Transformations of multiple scene objects fixed
- Reference system and exploration points flags fixed
- Possibility to set near- and farplane for an animationFile or per Snapshot 
- Set changed parameters into exampleFile (docs\Pro3D_ShortUserManual\Snapshot.json)

PR #573, opened 2025-11-18
Switch from FreeFly to Orbit should recompute the orbit center

haraldsteinlechner:
First heuristics, picking the center ray when switching to arcball works nicely (when there is no hit, we are still looking for a heuristics or user-feedback @Brunnhuber )

### DIFF
--- a/PRODUCT_RELEASE_NOTES.md
+++ b/PRODUCT_RELEASE_NOTES.md
@@ -1,3 +1,12 @@
+
+## 5.4.0
+- planets updates for local resSys in transforamtions
+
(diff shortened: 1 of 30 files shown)


haraldsteinlechner:
When swichting from Free Fly to Arcball, picking does not work initially. Resetting the arcball fixes this. however, if it cannot be fixed finding out why picking is borken would be vital, i suspect that when computing the arcball setttings from free fly (transition), some parameters are not taken into account when setting the scene.view.cameraview.

@Brunnhuber you are right into this right, can you take a quick look there as well?

PR #598, opened 2026-01-14
Features/geojsonexport test

sophie-sudoku:


Build.fsproj
```diff
@@ -12,5 +12,8 @@
     <Compile Include="Helpers.fs" />
     <Compile Include="Build.fs" />
   </ItemGroup>
+  <ItemGroup>
```
haraldsteinlechner: why does the build script require json parsing here? if this is really needed paket.dependencies should be used.

src/PRo3D.Base/Annotation/Exporters/GeoJSON.Export.fs
```diff
@@ -130,14 +128,25 @@ module GeoJSONExport =
...
 
+    let writeGeoJSONQGIS
+        (cooConfig : GeoJsonQGIS.CoordinateConfiguration)
+        (path        : string) 
+        (isSelected : Annotation -> bool)
+        (annotations : list<Annotation>) 
+        : unit = 
+
+        if path.IsEmpty() then 
+            ()
```
haraldsteinlechner: invariant of valid path should be checked before calling the function instead of silently doing nothing.

haraldsteinlechner: i think the better place would be where the action is raised.

src/PRo3D.Base/PRo3D.Base.fsproj
```diff
@@ -75,5 +76,8 @@
       <HintPath>..\..\lib\JR.Wrappers.dll</HintPath>
     </Reference>
   </ItemGroup>
+  <ItemGroup>
```
haraldsteinlechner: please add it to paket.dependencies

src/PRo3D.Core/Drawing/Drawing-App.fs
```diff
@@ -355,6 +358,28 @@ module DrawingApp =
         with e -> 
             Log.warn "[Drawing] exportGeoJson failed with %A" e
 
+    let exportGeoJsonQGIS
+        (cooConfig   : GeoJsonQGIS.CoordinateConfiguration)
+        (model       : DrawingModel) 
+        (path        : string) =
+
+        // export only visible annotations
```
haraldsteinlechner: i think this is duplicate code. i suggest creating a function which extracts the selected annotation

src/PRo3D.Core/Drawing/Drawing-App.fs
```diff
@@ -601,6 +626,26 @@ module DrawingApp =
             
                 model
 
+            | ExportAsGeoJSONQGIS_latlon path, _, _ ->     
+            
+                let planet = smallConfig.planet.Get(bigConfig).ToString()
```
haraldsteinlechner: regarding my other comment, please shortcut empty paths either here and add a warning. The best way would be to notify the user here that no path was selected and export won't happen, not to complicate things i think logging is sufficient though

src/Tests/SpiceTests.ipynb
```diff
@@ -0,0 +1,176 @@
+{
```
haraldsteinlechner, line 1: would be nice to clear the outputs

haraldsteinlechner (changes requested):
thanks sophie, i added some comments.

src/PRo3D.Base/Annotation/Exporters/GeoJSON.Export.fs
sophie-sudoku: the path is only empty if the user cancels the export, so it shooooould be ok if we do quiet quitting here

Build.fsproj
sophie-sudoku: done

src/PRo3D.Base/PRo3D.Base.fsproj
sophie-sudoku: done

src/Tests/SpiceTests.ipynb
sophie-sudoku: done

src/PRo3D.Core/Drawing/Drawing-App.fs
sophie-sudoku: done

src/PRo3D.Base/PRo3D.Base.fsproj
sophie-sudoku: ... edit: the build fails without this code in PRo3D.Base.fsproj (even though the dependency is added in paket.dependencies)

src/PRo3D.Core/Drawing/Drawing-App.fs
sophie-sudoku: quiet quitting seems not to be a problem i think (see comment above) but i added an error message now to the json file, if no reference frame is set, such that the user knows how to fix the problem

haraldsteinlechner (approved):

PR #600, opened 2026-02-03
Add rimfax manual

sophie-sudoku:
i did not add an .aux file for rimfax.tex - do we need it?


haraldsteinlechner:
thanks! and for noting the aux file. aux files should not be in version control anyways and should be ignored. i will remove it and append my change to your commit

haraldsteinlechner (approved):
just merged develop back, aux files were already removed in develop, i just changed the version

PR #619, opened 2026-06-08
expanded csv export of annotations and added text display in annotation ui

ThomasOrtner:
Added lon lat alt values to csv export of annotations. Further, this PR adds a way to display annotation information via the annotation text in the list ui element. 

If text is empty, it falls back to the standard behavior of displaying the geometry type, else a ui annotation item shows the annotation text. Both cases are overruled if semantic horizon information from the correlation panels is present.


sophie-sudoku:
1) the updated latlon code (ExportAsCsv) is currently not in use, is it still useful to add latlon here?
2) should we maybe merge this into release/6.0.0 as this is our most up-to-date branch (EDIT: i retargeted it)

PR #622, opened 2026-06-25
Add group default colors

sophie-sudoku:
add a color picker for annotation group default colors


Brunnhuber:
Button for color choosing besides Annotation type has to be removed to prevent user confusion.

Brunnhuber (commented):
Does what it should

PR #624, opened 2026-06-30
Add test protocol

sophie-sudoku:


haraldsteinlechner:
unfortunately we cannot add the heavyweight OPCs test data to the source code repository. this one needs a rework where the opcs are purged from the history i think, otherwise we have pull in huge binary blobs.

PR #629, opened 2026-07-27
Releases/6.0.0 bulk editing

sophie-sudoku:
Add bulk edit

grafik

worth noting: the values in the fields always belong to the first of the selected items... having values inside the fields makes the usage of the arrow keys more intuitive... 

* undo redo is implemented for changes
* a new toggle UI element is implement to set a value for toggles (such that all the selected items get the same value and are not just toggled)


src/PRo3D.Base/GuiEx.fs
```diff
@@ -96,6 +96,24 @@ module GuiEx =
     let iconCheckBox (dings : aval<bool>) action =
       iconToggle dings "check square outline icon" "square icon" action
 
+    // Like iconCheckBox, but the click emits an *absolute* target value (the negation of the
```
haraldsteinlechner, line 99: `dings` should be `predicate`

haraldsteinlechner (changes requested):

PR #631, opened 2026-07-28
Releases/6.0.0 rose diagram

sophie-sudoku:
Add rose diagram to annotation bulk edit


haraldsteinlechner:
what happens when a group contains thousands of annotations, could this be a problem? are the edges known or evaluated? If not a feature toggle would be useful.

sophie-sudoku:
> what happens when a group contains thousands of annotations, could this be a problem? are the edges known or evaluated? If not a feature toggle would be useful

a feature toggle for the developer or for the user? i.e. a button in the UI or a flag that the developer can easily set?

haraldsteinlechner:
Reviewed the head revision (06ab538). `RoseDiagram.fs` itself is fine — the concerns are all in how `ViewerGUI.fs` feeds it. I have not measured this, only read it, which is also why I'm asking for a test scene at the bottom.

## Adaptive graph is built per selected annotation

In `viewBulkAnnotationProperties`:

```fsharp
let angles : aval<list<float>> =
    let ids = selectedList |> List.map (fun ts -> ts.id)
    alist {
        let! usePoly = model.roseUsePolyline
        let! useDns  = model.roseUseDnS
        for annoId in ids do
            match! AMap.tryFind annoId annotations.flat with
            | Some (AdaptiveAnnotations a) ->
                let! geo = a.geometry
                ...
                let! az = AVal.bindAdaptiveOption a.dnsResults nan (...)
    }
    |> AList.toAVal |> AVal.map IndexList.toList
```

Each `match!` / `let!` inside the `for` compiles to a nested `AList.bind`, so per selected annotation this constructs roughly 5–8 adaptive nodes plus ~3 list readers. For a group `Select All` over N annotations that is O(N) graph nodes to produce a 16-bin histogram.

**1. `AMap.tryFind` called once per annotation.** FSharp.Data.Adaptive 1.2.x implements it as

```fsharp
/// Note that this operation should not be used extensively since its resulting
/// aval will be re-evaluated upon every change of the map.
let tryFind (key: 'K) (map: amap<'K, 'V>) = map.Content |> AVal.map (HashMap.tryFind key)
```

so all N lookups become separate outputs of `flat.Content`. `flat` is a `ChangeableModelMap`, so in-place edits to an annotation do not invalidate it — but structural changes do: finishing a new annotation, deleting one, import, scene load. Each of those marks all N lookups outdated, re-runs the `alist` and replaces the SVG.

**2. The two toggles are bound at the root of the `alist`.** `let! usePoly` / `let! useDns` sit above the `for`, so clicking either checkbox invalidates the outermost bind and disposes/reconstructs every per-annotation subtree. Same for the enclosing `adaptive { let! selected = annotations.selectedLeaves.Content ... }` — adding one annotation to the multi-selection rebuilds the whole panel and the whole rose graph.

**3. `AList.toAVal` discards the per-element tracking.** The result is collapsed to a whole list, then a fresh `float list`, then a fully rebuilt SVG subtree, so the fine-grained construction cost buys nothing.

**4. It stays live when hidden.** The panel is inside `GuiEx.accordion "Bulk Edit" "edit" false` (ViewerGUI.fs:1496) — collapsed by default, but accordion content is CSS-hidden rather than absent and the docked page stays live when the tab is not focused. The feature toggle does cover the off case correctly: `angles` is lazy and never pulled while `roseEnabled` is false, so nothing is constructed until the user switches it on.

**5. Rebuild churn.** Each rebuild orphans the previous node set. Outputs are weakly held so they are collectable, but until GC they remain in the output sets of `flat.Content`, `roseUsePolyline`, `roseUseDnS` and every annotation's `geometry` / `dnsResults`, and mark propagation walks them.

The code comment above the block states this "recomputes live and cheaply" — that holds for the evaluation, not for the construction.

## Suggested direction

A 16-bin histogram over the whole selection is a whole-collection aggregate, so whole-value semantics fit better than per-element tracking:

```fsharp
let angles =
    adaptive {
        let! groups  = annotations.Current      // AdaptiveGroupsModel.Current : aval<GroupsModel>
        let! usePoly = model.roseUsePolyline
        let! useDns  = model.roseUseDnS
        return
            ids |> List.choose (fun id ->
                match HashMap.tryFind id groups.flat with
                | Some (Leaf.Annotations a) when
                    (a.geometry = Geometry.Polyline && usePoly) ||
                    (a.geometry = Geometry.DnS      && useDns) ->
                        a.dnsResults |> Option.map (fun d -> d.dipAzimuth)
                | _ -> None)
    }
```

Three adaptive nodes instead of ~6N, no readers, no teardown/rebuild, one O(n) pure pass per invalidation. The trade-off is that `Current` invalidates on any annotation change rather than only structural ones. If that turns out to matter, the incremental alternative is a single `AMap.filter` + `AMap.toASet` (one reader on `flat`, incremental adds/removes) instead of N `tryFind`s. Either way, `usePoly` / `useDns` should not sit at the root position.

## Other observations

- `atan2 0 0 = 0`, so an isotropic or near-cancelling distribution draws the red mean line due north. The line is also drawn at full `radius` regardless of resultant length R, so a statistically weak mean renders identically to a tight one.
- `RoseDiagram.view` makes four passes over the list (`List.iter`, `List.length`, two `List.sumBy`); one fold covers all of it.
- `dnsResults` is computed from point count rather than geometry (`Drawing-App.fs:156`), so a Polyline can carry a dip azimuth. Worth confirming the Polyline toggle matches the intended geology semantics.
- No `docs/` page for the feature — CLAUDE.md asks for one in the same change.
- The diff carries unrelated `.g.fs` churn (field reordering in `SequencedBookmarks-Model.g.fs`, `updatePerFrame` in `Navigation-Model.g.fs` with no corresponding `-Model.fs` change). Adaptify non-determinism from the merge rather than a bug, but it adds diff noise.
- PR targets `releases/6.0.0` while current work is on `releases/6.1.0`.

## Request

Could you add a test scene to [PRo3D.Resources.TestData](https://github.com/pro3d-space/PRo3D.Resources.TestData) with a large set of DnS/Polyline annotations — ideally a few thousand in one group so `Select All` exercises the bulk path? That would let the scaling above be measured rather than reasoned about, and gives us a regression case for the bulk-edit panel generally.

haraldsteinlechner:
@sophie-sudoku please retarget to current develop or to releases/6.1.0 if no reason to keep it back from 6.1 track.

PR #665, opened 2026-08-13
fix(cross-section): don't clip against an unfilled attribute (Apple Silicon dark quads)

haraldsteinlechner:
## Problem

On Apple Silicon the OPC surface rendered a regular lattice of black holes across the terrain. Windows and Linux were unaffected.

The cause was `crossSectionClip` discarding fragments based on a per-vertex attribute that had never been filled.

`clippingEnabled` defaults to `true` and only means "clip if there is something to clip against", but nothing checked whether a cross-section actually existed. So the clip shader ran from the first frame of every scene and read `InsideOutsideV4`, which `Surface.Sg` binds as a **constant** vertex attribute (`SingleValueBuffer`) while no cross-section is defined. On Apple Silicon that value does not arrive — whole patches read back garbage instead of the bound zero, and roughly half of it is negative, so it was discarded.

## Fix

A `CrossSectionDefined` uniform, so the attribute is only read once it holds real data. Zero cost.

Substituting a real zero-filled `ArrayBuffer` also fixes it and was used to confirm the diagnosis, but would cost an extra `Patch.load` per patch on the no-cross-section path — i.e. almost always.

Whether the underlying fault is Apple's GL driver or how Aardvark's GL backend drives the constant-attribute path is **not** established; the comments say so rather than guessing. Reported upstream.

## Verification

Full 22-stage `surfaceEffect`, no environment overrides, same scene and camera that showed the artefact: clean. `contourLines` alone was rendered as a matched control and was clean, isolating the cause to this one stage.

## Also included — the tooling this was found with

- **`src/OpcViewer/ScreenshotViewer.fs`** — headless/interactive OPC renderer with an effect *ladder* that composes progressively more of the viewer's surface effect (`minimal` → `filter` → `normals` → `lit` → `color`), plus a CLI (`--dataset`/`--eye`/`--bearing`/`--pitch`/`--stack`/`--dump-glsl`). `--eye/--bearing/--pitch` take the numbers straight off PRo3D's HUD, so a bug report becomes one command.
- **`ViewerUtils.surfaceEffect` is now a named, dependency-aware stage list**, filterable via `PRO3D_SURFACE_EFFECT` / `_ADD` / `_DROP`. Unset changes nothing. `stageDependencies` drops consumers when their producer is dropped, so any subset is valid instead of throwing `Could not get attribute`.
- **`crossSectionDebug`** — paints the tested attribute instead of discarding on it (green = 0, red = negative, blue = positive). This is what turned the diagnosis from inference into observation.
- **`scripts/capture-surface-effect.sh`** — one unattended bisect round: launch on the most recent scene, screenshot, shut down.
- **`PRo3D.Base.OpcSurfaceShader`** — `stableTrafo` + `triangleSizeFilter` moved out of the viewer so the harness and the viewer share the same shaders rather than copies that drift.

## Docs

`docs/OpcViewer-Screenshot-Harness.md` (new) and a mechanism section in `docs/CrossSections.md`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,7 @@ testbed-output/
 !**/.yarn/sdks
 !**/.yarn/versions
 src/Tests/logs/
+
+# harness / bisect screenshot output
+capture-output/
+opcviewer-output/
--- a/docs/CrossSections.md
+++ b/docs/CrossSections.md
@@ -67,6 +67,55 @@ Next choose curtain settings and set up the curtain:
 
 ![alt text](images/crossSectionAndCurtain.png)
 
+## How clipping is implemented
+
+`Surface.Sg` writes a **signed distance to the cross-section polygon** into a per-vertex
+`InsideOutsideV4` attribute (negative inside), interpolated across triangles so the clip
+edge is smooth rather than snapping to mesh-edge midpoints. The `crossSectionClip`
+fragment stage in `ViewerUtils.surfaceEffect` discards fragments whose interpolated
+distance is negative.
+
+Two flags gate it, and they are not redundant:
+
+- **`CrossSectionClippingEnabled`** — the user's Clipping checkbox. Defaults to **true**
+  (`CrossSection-Model.fs`), so it only means "clip if there is something to clip against".
+- **`CrossSectionDefined`** — whether a cross-section actually exists.
+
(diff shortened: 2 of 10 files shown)


haraldsteinlechner:
not a self contained fix, will not merge this one.

Issue #667, opened 2026-08-13
Use SurfaceAttributes from OPCViewer.Base instead of PRo3D's duplicate opcx parser


haraldsteinlechner:
PRo3D carries its own copy of the `*.opcx` attribute layer parser:

- `PRo3D.Core.Surface.SurfaceUtils.SurfaceAttributes` — `src/PRo3D.Core/Surface/SurfaceApp.fs`
- called from `SceneLoader.addSurfaceAttributes` — `src/PRo3D.Viewer/Scene.fs`

It duplicates `OpcViewer.Base.Attributes.SurfaceAttributes` from the `OPCViewer.Base` package almost line for line (`get`, `parseMap`, `parseTexture`, `parseLayer`, `layers`, `read`). The only real difference is that PRo3D's `ScalarLayer`/`TextureLayer` carry a `version` field for scene persistence.

The duplication just cost us: multi-channel `ChannelsDefinedRange`/`ChannelsActualRange` failed to parse, so OPCs with vector-valued attribute layers (normals, gravity vectors, lon/lat/radius) could not be imported without hand-editing the `*.opcx`. The bug existed in **both** copies and had to be fixed twice:

- upstream: aardvark-platform/OPCViewer.Base#1 / aardvark-platform/OPCViewer.Base#2
- here: fixed locally in `SurfaceUtils.SurfaceAttributes.parseChannelRange`, covered by `OpcSidecarTests.opcx attribute layers` in `src/Tests/OpcSidecarTests.fs`

### What to do

Once the upstream fix is released (>= 1.11.1), drop PRo3D's copy and call `OpcViewer.Base.Attributes.SurfaceAttributes` instead.

The one thing to sort out is the layer types. Upstream returns *its* `ScalarLayer`/`TextureLayer`, PRo3D needs its own versioned ones, so this needs either a thin mapping layer in PRo3D or an upstream parser that is generic over the layer record. Worth deciding which before touching it — a mapping layer is the smaller change but keeps a second definition of the model around.

Keep the tests in `OpcSidecarTests` either way; they exercise the behaviour PRo3D depends on regardless of where the parser lives.

PR #669, opened 2026-08-13
Read OPC attribute layers from per-vertex aara data, show them under the 3D cursor

### LINKED ISSUE #668: Extract OPC attribute layers from per-vertex aara data, and show them under the 3D cursor
Newer OPC exports (HERA `AARA_Textures`) ship their attribute layers twice: as texture layers under `Images/<Layer>/`, and additionally as **per-vertex `*.aara` grids** inside each patch directory, listed in `patch.xml`'s `<Attributes>`.

Profile extraction currently only samples the textures, which costs a full image decode per layer per sample. Reading the per-vertex form instead needs three small random-access reads per layer, so it is fast enough to run interactively.

### Wanted

* Read per-vertex attribute layers and use them for profile extraction, keeping texture sampling as the fallback for layers they do not cover.
* Show the values under the 3D cursor live, as the testbed for the extraction. The preview cursor should be on by default.
* Read the `*.opc.json` sidecar that ships next to the `*.opcx` — DEM reference model, and the DSKBRIEF summary of the source `*.bds` shape model for DSK-derived OPCs.

### Traps found while implementing

* The per-vertex grid is **smaller** than the position grid — the positions carry a symmetric skirt whose width shrinks with the hierarchy level (1032/1028/1026 against a constant 1024). `off = (posSize - attrSize) / 2`.
* Attribute **textures store values normalised** into the layer's `ChannelsDefinedRange`, while per-vertex layers hold physical values. The existing profile CSV export was therefore emitting `[0,1]` numbers, and nodata (`-99999`) as a real value.
* Multi-channel `ChannelsDefinedRange`/`ChannelsActualRange` could not be parsed at all, so such OPCs needed a hand-edited `*.opcx` to import (see #667 and aardvark-platform/OPCViewer.Base#2).
* `buildPatchInfoLookup` keyed patches case-sensitively while `OpcPaths.Patches_DirAbsPath` and the persisted kd-trees can spell `patches`/`Patches` differently, so extraction silently found nothing on affected datasets.

Data: `C:\pro3ddata\HERA\OPCUpdate\AARA_Textures` (Dimorphos, per-vertex layers) and `...\BDS_Metadata` (Deimos, DSK-derived sidecar).

haraldsteinlechner:
Closes #668. Related: #667 (drop PRo3D's duplicate opcx parser), aardvark-platform/OPCViewer.Base#2 (same parser bug upstream).

Newer OPC exports ship their attribute layers twice: as texture layers under `Images/<Layer>/`, and additionally as **per-vertex `*.aara` grids** in each patch directory. Reading the per-vertex form needs three small random-access reads per layer instead of a full image decode, which makes attribute extraction fast enough to run interactively.

Docs: [`docs/VertexAttributes.md`](docs/VertexAttributes.md).

## The grid offset

The attribute grid is **smaller** than the position grid — the positions carry a symmetric skirt whose width shrinks with the hierarchy level, because it is a constant number of source DEM pixels:

| level | positions | attributes | offset |
|-------|-----------|------------|--------|
| 0     | 1032²     | 1024²      | 4      |
| 1     | 1028²     | 1024²      | 2      |
| 2     | 1026²     | 1024²      | 1      |

```
attributeIndex(x, y) = positionIndex(x + off, y + off),   off = (posSize - attrSize) / 2
```

Established empirically: `LonLatRad`'s third channel is the vertex radius in metres, so it must equal `|Local2Global · XYZ_Local|`. Across all 9 patches of `g_01960mm_spc_dtm_dimo_0000n00000_v003` the centred offset agrees to float32 round-off (median 2·10⁻⁶ m); **every other offset is off by ≥ 0.19 m**. The same invariant guards the code as a test.

## Three problems the comparison surfaced in the existing texture path

1. **Texture samples were normalised, not physical.** Attribute textures store each layer scaled into its `ChannelsDefinedRange`; per-vertex layers hold physical values. The profile CSV was therefore exporting `[0,1]` numbers, and the nodata sentinel `-99999` as if it were a measurement. Samples are now mapped back onto the layer's range and nodata is dropped.
2. **Most layers were unreachable.** The loop counted `DiffuseColorNWeights` entries as textures and capped at `numTextures - 1`, so it only ever reached 3 of the 8 layers on this dataset.
3. **The patch lookup silently missed everything on some datasets.** `buildPatchInfoLookup` keyed on the full positions path case-sensitively, but `OpcPaths.Patches_DirAbsPath` probes `"patches"` before `"Patches"` (and `Directory.Exists` is case-insensitive on Windows) while persisted kd-trees may carry either spelling. Now case-insensitive with separator normalisation.

## Under Cursor readout

The 3D preview cursor is **on by default** and its readout is the *Under Cursor* section of the **Config** panel, below *Screenshots*. It shows the surface, patch and every per-vertex layer under the mouse (hold `CTRL`).

Per-vertex data only — the texture fallback is one image decode per layer and cannot run per mouse move, so a surface without per-vertex layers says so rather than stalling. Extraction runs on the existing background picking thread.

## Memory

The triangle-to-grid mapping is now one int per quad instead of an `int[3]` per triangle — **4 MB instead of ~100 MB per 1032² patch** — and the cache is bounded to 32 patches, because the cursor pulls in a new patch whenever it crosses a patch boundary. Caches are dropped when a surface is removed. ZIP-backed OPCs cannot seek, so their payloads are held in memory instead, bounded at 512 MB.

## Multi-channel `*.opcx` ranges

For multi-channel `Map` layers ExportGpc writes one range per channel:

```xml
<ChannelsDefinedRange>[[-0.000039896, 0.000040985], [-0.000046144, 0.000046183], [-0.000050788, 0.000050736]]</ChannelsDefinedRange>
```

`Range1d.Parse` only understands `[min, max]`, so these OPCs could not be imported without hand-editing the `*.opcx`. Both forms parse now. The **first** channel's range is kept — that is the channel the false-colour legend and the de-normalisation above refer to (attribute textures are read via `ChannelWithIndex 0`); unioning the channels would widen the `Gravity` range by 25% and skew every de-normalised value.

## `*.opc.json` sidecar

`OpcMetadata` reads the sidecar next to the `*.opcx` and logs it at import (not persisted into the scene): product provenance, the DEM reference model, and for DSK-derived OPCs the DSKBRIEF summary of the source `*.bds`. `DemSphere` and `DemEllipsoid` have different schemas (single `Axis` vs `AxisX/Y/Z` + `Radii`) and both are read.

## Verification

Built and run against `C:\pro3ddata\HERA\OPCUpdate`:

* Solution builds; `OpcSidecar` 10/10, `ProfileAttributeExtraction` 2/2 (4 skipped — they need `Dimorphos_DRACO1`, absent here). Data-backed tests self-skip; point them with `PRO3D_AARA_OPC` / `PRO3D_BDS_OPC`.
* **Grid offset**: radius invariant validated at 6 hits across patches at all three levels.
* **De-normalisation**: de-normalised texture samples match per-vertex values to ≤ 7% of a layer's range for smooth layers; without it the deviation is on the order of a whole range.
* **Unpatched `*.opcx`**: fed the original ExportGpc file, the parser yields ranges **identical to the hand-patched one**, layer for layer. A copy of the OPC with the original sidecar sits next to the patched one for end-to-end checks.
* **In the app**: imported the un-hacked OPC, `7 per-vertex layers (LonLatRad, Normal, Gravity, Magnitude, Potential, Elevation, Slope)`, no lookup warnings, `[OpcMetadata]` block logged for both the `DemSphere` and `DemEllipsoid` exports.

Not verified: the *Under Cursor* panel has not been screenshotted, only exercised by hand.

### DIFF
--- a/ai/RENDERING.md
+++ b/ai/RENDERING.md
@@ -28,6 +28,7 @@ OPC is the patch-based, hierarchical terrain format PRo3D is built around. An OP
 - **Loading**: `Aardvark.Data.Opc` parses the hierarchy and patch files; PRo3D discovers/validates dataset folders in `src/PRo3D.Core/Surface/` (folder discovery, `isOpcFolder`/`isSurfaceFolder`).
 - **LoD rendering**: `Aardvark.GeoSpatial.Opc` provides `Sg.patchLod` (and friends), which builds an adaptive scene-graph node that streams and swaps patch levels based on screen-space error / distance metrics.
 - **`opc-tool`** (`src/opc-tool/`) is the offline companion CLI: validate a dataset, resize/convert patch textures (to `.dds`), and **pre-build KdTrees** for picking.
+- **Attribute layers** (elevation, slope, gravity, …) come in two forms. Every OPC has them as *texture layers* under `Images/<Layer>/`, declared in the dataset's `*.opcx`. Newer exports additionally ship them as *per-vertex* `.aara` grids inside each patch directory, listed in `patch.xml`'s `<Attributes>`. Reading values at a picked point prefers the per-vertex form (three small random-access reads per layer) and falls back to decoding the texture. **Beware two traps**: the per-vertex grid is *smaller* than the position grid (a centred skirt, `off = (posSize − attrSize) / 2`), and texture layers store values *normalised into the layer's `ChannelsDefinedRange`* while per-vertex layers hold physical values. See [docs/VertexAttributes.md](../docs/VertexAttributes.md) and `src/PRo3D.Core/VertexAttributes.fs`.
 
 ---
 
--- a/docs/VertexAttributes.md
+++ b/docs/VertexAttributes.md
@@ -1,0 +1,201 @@
+﻿# Per-Vertex Attribute Layers
+
+Newer OPC exports store their attribute layers twice: as **texture layers** under
+`Images/<Layer>/` and, additionally, as **per-vertex `*.aara` grids** inside each patch
+directory. PRo3D reads the per-vertex form wherever it exists — it is far cheaper than
+decoding an image per layer per sample — and falls back to texture sampling otherwise.
+
+Two features use this:
+
+* **Profile extraction** — attribute values along an annotation, exported to CSV.
+* **The 3D cursor readout** — the *Under Cursor* section of the **Config** panel, which
+  shows the values under the mouse pointer live. This is also the testbed for the
+  extraction itself.
+
+## What the data looks like
+
+A patch directory of such an OPC (HERA Dimorphos, `AARA_Textures` export):
+
+```
+Patches/0_0_2/
+  XYZ_Local.aara              positions (V3f, 1032 x 1032)
+  Earth8K_Coordinates.aara    texture coordinates (V2f, 1032 x 1032)
+  Earth8K_Weights.aara        texture blend weights (float, 1032 x 1032)
+  LonLatRad.aara              V3f, 1024 x 1024   <-- attribute layers
+  Normal.aara                 V3f, 1024 x 1024
+  Gravity.aara                V3f, 1024 x 1024
+  Magnitude.aara              float, 1024 x 1024
+  Potential.aara              float, 1024 x 1024
+  Elevation.aara              float, 1024 x 1024
+  Slope.aara                  float, 1024 x 1024
+  patch.xml
+```
+
+`patch.xml` lists them in its `<Attributes>` element; the layer name is the file's base
+name. Only layers whose resolution matches the geometry can be exported this way, so an
+OPC may well ship a subset — or none at all.
+
+### The attribute grid is smaller than the position grid
+
+The position grid carries a skirt around the attribute grid, and its width shrinks with
+the hierarchy level because it is a constant number of *source DEM* pixels:
+
+| level | positions | attributes | offset |
+|-------|-----------|------------|--------|
+| 0     | 1032²     | 1024²      | 4      |
+| 1     | 1028²     | 1024²      | 2      |
+| 2     | 1026²     | 1024²      | 1      |
+
+So a vertex at position grid `(x, y)` maps to attribute index
+
+```
+attributeIndex = (y - off) * attrWidth + (x - off),   off = (posSize - attrSize) / 2
+```
+
+This was established empirically, by comparing the third `LonLatRad` channel (the vertex
+radius, in metres) against `|Local2Global · XYZ_Local|` for every patch of
+`g_01960mm_spc_dtm_dimo_0000n00000_v003`. The centred offset agrees to float32 round-off
+(median error 2·10⁻⁶ m); every other offset is off by at least 0.19 m. The same invariant
+guards the code — see `per-vertex layers are physically consistent` in
+`src/Tests/ProfileAttributeExtractionTest.fs`.
+
+## How a value is sampled
+
+1. Pick a surface: the KdTree hit gives a triangle index into the patch's triangle set.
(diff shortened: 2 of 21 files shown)


haraldsteinlechner:
Reviewed with a sub-agent. It found one real defect in the load-bearing assumption, which is now fixed and pinned by tests, plus four smaller ones. All in `a7bc528`.

## The mapping followed the wrong triangle numbering

`hit.SetObject.Index` indexes the object set built by `DebugKdTreesX.loadTriangles'` = `PRo3DCSharp.ComputeIndexArray` + `getTriangleSet`, and **that path does not compact**. `ComputeIndexArray` writes into a pre-allocated array and does `k += 6; continue` for an invalid quad, so the quad becomes the degenerate triangle `(positions[0], positions[0], positions[0])` — which `getTriangleSet`'s NaN filter only drops when `positions[0]` is *itself* NaN.

The mapping enumerated valid quads only. So on a patch with a valid vertex 0 and any invalid quads, every lookup was displaced by the number of invalid quads preceding the hit — reading elevations and slopes off the wrong vertices, plausibly and silently.

It passed the radius invariant because the Dimorphos patches have an all-NaN border row, so `positions[0]` is NaN, the fillers get dropped, and the two orderings coincide. A patch with interior holes and a valid corner would not have been so kind. (`reports/TriangleSetOptimization.md` records 16,432 skipped quads on patch `0_0_1` — about 16 grid rows of displacement.)

`computeTriangleQuadStarts` now mirrors that predicate exactly: valid quads get their index, invalid ones get `-1` when the fillers survive and are skipped when they do not. Two tests in `TriangleSetTests` check the mapping against the *actual* `ComputeIndexArray + getTriangleSet` output for both cases — verified failing before the fix (`13 passed, 1 failed`, on `mapping must cover exactly the triangles the kd-tree holds`) and passing after.

## Also fixed

- **`buildPatchInfoLookup` raced.** Unsynchronized `Dictionary`, now reached from both the background picking thread and profile export on the update thread. Locked. `clearCaches` also drops it — it was left behind despite the doc claiming all caches were cleared.
- **`VertexAttributes` lock discipline.** Two locks for three pieces of state, and the eviction path cleared `layerCache` while holding only the `fileCache` lock. One lock now guards all three.
- **Layer-name comparison was case-sensitive.** Texture names come from the `Images/<Layer>` folder, per-vertex names from the `*.aara` base name; a spelling difference let the texture value through and, in `extractProfile`, *override* the per-vertex one — inverting the stated priority.
- **`attributeIndex` accepted offsets it cannot centre.** An odd size difference (asymmetric skirt) or a negative one silently shifted every sample by a vertex. Refused with a warning now, falling back to texture sampling.
- **Stale read-out.** A preview pick with no hit left the previous point's values on screen as if current. Cleared now, and the empty state no longer claims the surface has no layers when the cause is a hit on the skirt.

## Not acted on

- The reviewer objects that `OpcMetadata` (~550 lines incl. tests) and the `showPreviewIntersection` default flip are separable features. Both were explicitly requested alongside this work, so they stay here deliberately.
- `AnnotationQuery.handlePatch` reads the same `*.aara` layers and indexes them with the **position** grid index directly, i.e. assuming both grids are the same size — the assumption this PR disproves. On these OPCs that path throws or reads the wrong vertex. Out of scope; filed separately as a follow-up so `VertexAttributes.attributeIndex` becomes the single source of that knowledge.

Solution builds; `TriangleSet` 14/14, `OpcSidecar` 10/10, `ProfileAttributeExtraction` 2/2.

haraldsteinlechner:
Follow-up filed as #670.

PR #673, opened 2026-08-13
fix(snapshots): match the offscreen framebuffer signature to the app's

haraldsteinlechner:
Every snapshot rendered black. A 182-frame sequenced-bookmark animation produced 181 identical 3150-byte PNGs, and the run performed exactly **one** texture load â€” nothing was ever drawn.

## Root cause

The runtime rejected each prepared render command:

```
ERROR: [GL] Prepared command has framebuffer signature
  { Samples = 1 ... DepthStencilAttachment = Some Depth24Stencil8 }
but expected
  { Samples = 8 ... DepthStencilAttachment = Some Depth24Stencil8 }
ERROR: [SNAPSHOT] Could not render image 000000_Bookmark_0.png
```

A render object may only be drawn into a framebuffer whose signature matches the one it was prepared against. The scene graph is prepared against the app signature in `PRo3D.Snapshots/Program.fs`, but each snapshot path built its own offscreen target from hardcoded values â€” and all three had drifted apart:

| | samples | depth |
|---|---|---|
| `Program.fs:100` (app signature) | *no argument* â†’ 1 | `Depth24Stencil8` |
| `executeBookmarkAnimation` | 8 | `Depth24Stencil8` |
| `executeCameraAnimation` | 1 | `DepthComponent32f` |

So **bookmark** animations failed on the sample count, and **camera** animations failed on the depth format. The camera path is not a side case â€” `executeAnimation` routes `PanoramaCollection` through it, and `SnapshotAnimation.fs:313` falls back to a `CameraAnimation` when there are no sequenced bookmarks. All three snapshot animation kinds were affected.

## Fix

One source of truth, since three hardcoded copies had already drifted twice:

```fsharp
module SnapshotFramebuffer =
    [<Literal>]
    let samples = 8
    let depthStencilFormat = TextureFormat.Depth24Stencil8
    let attachments = [ DefaultSemantic.Colors, TextureFormat.Rgba8
                        DefaultSemantic.DepthStencil, depthStencilFormat ]
    let createSignature (runtime : IRuntime) =
        runtime.CreateFramebufferSignature(attachments, samples)
```

Both snapshot paths and the app signature now build from it. Samples stay at **8**, so snapshots keep their multisampling.

## Testing

Against a real Jezero scene (9 OPC surfaces) and the `batchRendering.json` the viewer emits, on both paths, zero signature errors:

| Path | Before | After |
|---|---|---|
| Bookmark animation | 181 Ã— 3150 B, all black | **182** frames, 1.32â€“2.04 MB, all distinct |
| Camera / panorama animation | 4 Ã— 3150 B, all black | 5 frames, 1.32â€“1.38 MB, all distinct |

The bookmark run gains a frame because `000000` previously failed outright rather than rendering black. Output was compared against the same view in the viewer and matches.

Reproduced identically on the rc1 standalone package and on a local Debug build beforehand, so this is a `releases/6.0.0` code bug rather than something about the packaging.

## Reviewer note

The camera path's depth attachment changes from `DepthComponent32f` to `Depth24Stencil8`. That buffer feeds `renderDepth` (`--renderDepth`), which is a live CLI flag — not dead code, despite the stale *"not in use"* comment at `SnapshotApp.fs:74`.

Tested on this branch: `--renderDepth` runs clean and writes a `_depth.exr` per frame. Depth *content* is separately broken — five different camera positions give five byte-identical depth files — but that is pre-existing and unrelated: rebuilding with `samples = 1` yields the same md5, so it does not depend on the sample count. Filed as #674.

So this change does not regress anything that currently works.

Needed for **rc2**: 6.0.0-rc1 ships with all snapshot animation kinds rendering black.

### DIFF
--- a/src/PRo3D.SimulatedViews/Snapshots/SnapshotApp.fs
+++ b/src/PRo3D.SimulatedViews/Snapshots/SnapshotApp.fs
@@ -21,6 +21,28 @@ open Chiron
 
 
 
+/// The framebuffer the snapshots are rendered into has to have the *same* signature the
+/// scene graph's render objects were prepared against, otherwise the runtime rejects the
+/// prepared command ("Prepared command has framebuffer signature ... but expected ...")
+/// and every frame comes out black. Both the offscreen targets here and the app signature
+/// in PRo3D.Snapshots/Program.fs are built from these values so they cannot drift apart.
+module SnapshotFramebuffer =
+    open Aardvark.Rendering
+
+    [<Literal>]
+    let samples = 8
+
+    let depthStencilFormat = TextureFormat.Depth24Stencil8
+
+    let attachments =
+        [
+            DefaultSemantic.Colors, TextureFormat.Rgba8
+            DefaultSemantic.DepthStencil, depthStencilFormat
+        ]
+
+    let createSignature (runtime : IRuntime) =
+        runtime.CreateFramebufferSignature(attachments, samples)
+
 type NearFarRecalculation =
   | Both
   | FarPlane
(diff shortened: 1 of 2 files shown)


haraldsteinlechner:
Correcting the reviewer note in the description — I checked `renderDepth` properly instead of trusting the code comment.

**It is not dead code.** It is reachable via the `--renderDepth` CLI flag (`CommandLine.fs:93`), wired through `StartupArgs` → `SnapshotGenerator` → `SnapshotApp.renderDepth` (used at `SnapshotApp.fs:210/235/270`), and `Section16_CommandLine.fs:41` asserts the flag parses. The *"not in use"* comment refers to the Mars-DL project that motivated it, not to the code path.

**Tested with this branch:** `--renderDepth` runs clean — no errors, a `_depth.exr` written per frame alongside the colour PNG.

**But depth output has a separate, pre-existing bug:** across five different camera positions producing five visibly different colour images, all five `_depth.exr` files are byte-identical.

To rule out multisampling as the cause I rebuilt with `SnapshotFramebuffer.samples = 1` and re-ran. The depth EXRs are still byte-identical — and carry *the same md5* (`039bdbacf3c2753ce2cd073525e1f929`) as the 8-sample run. The depth content does not depend on the sample count at all, so this change neither causes nor fixes it. Worth its own issue.

So the depth-format change in this PR is safe in the sense that matters here: it does not regress anything that currently works.

PR #680, opened 2026-08-14
Fix axis 4P ellipse

sophie-sudoku:


haraldsteinlechner:
Note for the record: this was merged into `releases/6.0.0`, which sits at the `v6.0.0` tag and is not the active development line. The `v6.0.0` tag was unaffected — the released build is unchanged.

The fix has been forward-ported and is now on `releases/6.1.0` via #696 (merged as `8b1045cc`, clean merge of the same branch, no conflicts).

`releases/6.0.0` has since been locked read-only to prevent this happening again. It still carries the three commits from this PR past the `v6.0.0` tag; harmless, since the tag defines the release.

PR #689, opened 2026-08-14
Boolean operations on annotations: union and cut, geometry-first with property-based tests

haraldsteinlechner:
Boolean operations on annotations: union of selected annotations and cutting along a terrain-drawn polyline stroke, built geometry-first with property-based tests.

## Geometry (`PRo3D.Base.Geometry`)

- `PolyRegion2d.fs`: vendored attributed `PolyRegion<'a>` from aardvark.base `0420fd19` (implements aardvark-platform/aardvark.base#100), verbatim apart from the namespace. Delete and switch back to `Aardvark.Geometry` once a release carries that commit.
- `RegionOps.fs`: `cut` along a polyline stroke (two-point stroke = straight cut; only the stroke's ends must lie outside the region) and `merge`, on regions whose vertices carry their world position through the attribute channel. Cut lines are nudged off region vertices — LibTess drops tangent slivers when a stroke passes exactly through vertices (found by fixture replay: 14% of a U-shape's area vanished).
- `AnnotationRegionOps.fs`: annotation-level `union`/`cut` in a 2D chart (plane fitted over all operands' points). Surviving vertices keep their exact world position; invented vertices are terrain-projected via a raycast parameter. After a cut, blended attributes are contaminated by the synthetic side polygon, so survivors are matched by chart position and other vertices lifted through `chart.toWorld`. Union components smaller than the smallest operand are self-intersection artifacts and dropped (every legitimate component contains a whole operand); substantial holes refuse with a typed `Refusal`.

## Viewer

- **Union**: select ≥ 2 annotations, press the union icon in the annotations panel. One annotation per component, metadata from the first operand in tree order, measurements recomputed, single undo step. Holes refuse with a message.
- **Cut**: `Interactions.CutAnnotation` — select one annotation, Ctrl+click a stroke; the stroke colours green/red by a live dry-run of the cut. Enter applies (one closed polygon per piece), Backspace/Escape edit/discard, refused cuts keep the stroke. Rendering reuses the working-annotation pipeline; result rings are stored closed like drawn polygons.
- Both operations enrich their message with the sky-direction kd-tree raycast in the viewer, the same pattern as `AddPointAdv`.
- `GroupsApp.removeLeafById`: path-independent leaf removal (viewport picks store the root path in `TreeSelection` regardless of the leaf's group, making path-based removal unreliable).

## Testing

- `RegionInvariants.fs`: cut/merge/round-trip invariants stated once, shared by the example corpus, FsCheck generators (100 cases each), and fixture replay.
- `PRo3D.GeometryLab`: 2D SVG app driving the real `RegionOps` (draw, polyline cut, merge, undo/redo); exports `.region` fixtures to `src/Tests/data/regions/`, each replayed through the invariants — one test per file.
- `AnnotationRegionOpsTests.fs`: chart round-trips, attribute fidelity, refusals, and a regression with the Mars-scale rings from the first real-scene union failure.
- `Section20_BooleanOperations.fs`: union and cut driven through the real `DrawingApp.update` (areas, component explosion, undo/redo, dry-run colours, refusals). No GL.
- Suite: 316 passed, 0 failed. Manual verification on a Jezero OPC scene: union and cut both confirmed working.

## Docs

`docs/AnnotationBooleanOps.md`, `docs/GeometryLab.md`; design records in `plans/booleanOperations.md`, `plans/viewerIntegration.md`, `plans/testingStrategy.md`.


haraldsteinlechner:
integrated in 6.1.0

PR #692, opened 2026-08-19
feat(tool): unproject — image pixel coordinates to body-fixed surface coordinates

haraldsteinlechner:
Adds a third verb to `pro3d-tool`. Stacked on #691, which it takes as its base.

## Why

A Hera request: convert lists of image pixel coordinates — centroids of features picked in AFC and CubeSat-camera images — into body-fixed surface coordinates, and re-derive them whenever a new SPICE or shape-model version lands (roughly monthly). The pixel coordinates never change; their projection onto the surface does.

Wanted for the ESOC training in September 2026, at the latest before arrival in November.

## What it does

```
pro3d-tool unproject --opc <body-opc> --images <folder> --input centroids.csv --out out.csv
```

Each row names an image and a pixel in it. The pixel is unprojected through that image's instrument camera, intersected with the shape model, and the surface point is written next to the row it came from.

what unproject does

Pure CPU — no GPU, no display, like `kdtree` and unlike `sun-angles`. The ray is built and intersected in `double` in the body-fixed frame, so nothing goes through a `float32` shader.

Output keeps one row per input row, in input order, with the input's own columns carried through:

```
image,x,y,label,status,x_m,y_m,z_m,lat_deg,lon_deg,alt_m,range_m
ASP_..._NIR1_0.tif,320,256,crater_rim,ok,388.6271,-81.4280,83.9454,11.9374,11.8338,-3.6572,9701.699
ASP_..._NIR1_0.tif,5,5,off_the_limb,no-hit,,,,,,,
```

plus one column per per-vertex attribute layer the OPC carries — on the HERA AARA exports that is `Slope`, `Potential`, `Elevation`, `Magnitude`, `Normal`, `Gravity`, `LonLatRad`, which is the set the request asked for.

A pixel that misses the body is `no-hit` rather than dropped, and does not fail the run. An unresolvable image or a malformed row does.

## Shared, not copied

Following the argument in e602f7aa:

- `InstrumentObservation.pixelRay` / `projectToPixel` and `PixelConvention`, next to the `projectorCamera` they belong to
- `PRo3D.Core.Surface.HeadlessPicking` — OPC ray casting without a viewer, plus per-vertex attribute sampling at a hit
- `PRo3D.Tool.Spice` — the `PRO3D_SPICE_KERNELS` resolution `sun-angles` had privately, so both verbs now fail with the same message

## Pixel convention

How an input file addresses its pixels cannot be inferred from the imagery or its metadata, so it is **declared** with `--pixel-convention image|fits` and echoed in the log. Coordinates are floats throughout; pixel *k* covers `[k-0.5, k+0.5]`.

image versus fits addressing

The observing spacecraft is derived per image from its instrument, so one input file may mix AFC (Hera) and ASPECT (Milani) images; `--observer` overrides it.

## Testing

14 tests run in CI with no data or kernels: the pixel algebra, and the input/output tables — separator detection, header auto-detection, extra columns surviving, malformed rows, 1:1 output ordering, and numbers written invariant under a comma-decimal culture.

The test worth pointing at needs data and self-skips in CI: the OPC's own per-vertex `LonLatRad` layer, baked in at export, checked against lat/lon/alt derived by unprojecting a pixel, intersecting the kd-trees and running the hit through `CooTransformation`. The two sides share no step, so an error anywhere in the chain breaks the agreement. On the HERA Dimorphos AARA export they agree to **1e-4 degrees and 2e-4 m**.

`scripts/test-unproject.{sh,cmd}` runs the verb end to end, taking `PRO3D_TEST_DATA` and `PRO3D_SPICE_KERNELS`.

Full kernel-free suite: 245 passed, 4 ignored, 0 failed.

## Also here

- `docs/Pro3DTool-Unproject.md`, plus rows in the verb tables of `docs/Pro3DTool.md` and the tool README
- `docs/VertexAttributes.md` gains a note that `LonLatRad` is stored in **gradians** (`lon x 10/9`, `(lat + 90) x 10/9`, radius in metres) — established by the cross-check above and recorded nowhere until now
- `scripts/make-unproject-figures.py` regenerates both figures; the overview one is a real run of the tool
- tool bumped to **0.2.0**, so `Pack`/`Push` reship it

## Not done

CI does not yet have the SPICE kernels or test data, so the data-backed tests only run locally. That is being handled separately.


haraldsteinlechner:
## Self-review

Reviewed the diff as a reviewer rather than re-reading it as the author. Findings ordered by how likely they are to bite, with a note on whether I verified or inferred each.

### 1. No CSV quoting, on input or output — *verified*

`splitOn` is `line.Split ','` and `writeTable` is `String.Join(delimiter, cells)`. Neither understands RFC 4180 quoting. Probed it:

```
in:   ASP_..._NIR1_0.tif,320,256,"boulder, large"
out:  ASP_..._NIR1_0.tif,320,256,"boulder,large",ok,388.6271,...
```

The geometry is right and the column count survives — because the row is split into N cells and rejoined with the same delimiter — but the label lost its space. Worse case is a quoted delimiter *before* the coordinates: `"ASP, weird.tif",320,256` parses `"ASP` as the image and ` weird.tif"` as x, so the row becomes `bad-input`.

Acceptable for the centroid lists this is aimed at, and it fails loudly in the bad case rather than producing wrong coordinates. But "extra columns are carried through untouched" is documented, and that is not quite true. Worth either honouring quotes or narrowing the doc claim.

### 2. Nothing is written when no image resolves — *verified by reading*

The output contract is "one row per input row, in input order". If *no* image resolves — a mistyped `--images`, say — `run` exits 1 at the `firstOk` check before `writeTable`, so no file appears at all. Every other failure mode still writes the row with a status. Asymmetric; a table of `no-image` rows would be more useful and more consistent.

### 3. A broken patch logs once per ray — *inferred*

`HeadlessPicking.intersectAll` catches per-patch intersection failures and `Log.warn`s. That is right for one bad patch, but the warning is inside the per-ray loop, so a systematically unreadable patch emits one line per input row. A 10k-row job would drown in it. Wants deduplicating by bounding box.

### 4. Sidecar lookup is O(N×M) — *inferred, pre-existing*

`InstrumentMetadata.buildMbiIndex` re-parses **every** `*.mbi.json` in the folder on each `tryParseMetadataForImagePath` call, i.e. once per distinct image. Fine for the handful here; a folder of hundreds of images against hundreds of sidecars will be noticeably slow. It lives in `InstrumentMetadata` and is shared, so not for this PR, but this verb is the first caller likely to hit it at scale.

### 5. `--config` relative paths resolve against the CWD — *inferred*

Paths inside a config file are used as-is, so a config with relative paths only works from the directory it was written for. Resolving them relative to the config file would be less surprising.

### 6. The `.tif` must exist although it is never read — *verified*

Zero pixel-reading calls in the verb; it needs only the sidecar metadata. `resolveImage` still requires the file. This is shared with `sun-angles`, which *does* read pixels, so I left it alone and documented it — but it means you cannot unproject coordinates for imagery you do not hold locally, even though the geometry is entirely in the sidecar.

### 7. Input coordinates are not range-checked — *inferred*

`x = 5000` on a 640-wide image yields a ray outside the frustum and a quiet `no-hit`, indistinguishable from a genuine off-limb pixel. An out-of-frame coordinate is almost always a wrong-file or wrong-convention mistake and could say so. (I had a `pixelInImage` helper for this and cut it as speculative; on reflection the check has a real use.)

### 8. Scope dropped from the plan — *intentional, recording it*

`--config` was to accept an instrument FOV table so an unlisted CubeSat camera could be used without a code change. Not implemented. The docs do not promise it and the caveats do say an instrument outside the table cannot be used, so there is no mismatch — but the original request mentioned CubeSat cameras generally, and only ASPECT NIR1 is covered today.

### What I am confident in

The geometry is checked against something this code has no hand in producing — the shape model's own `LonLatRad` layer — and agrees to 1e-4 degrees and 2e-4 m across every sampled hit. That covers the ray construction, the kd-tree intersection, the barycentric attribute sampling and the coordinate conversion in one assertion. Sanity figures independently agree: hit radii land at 406–426 m against Didymos's ~409 m mean radius, and `range_m` ≈ 9.70 km against the sidecar's 10.21 km to body centre.

Full kernel-free suite is green (245 passed, 4 ignored), and the 14 new tests need no data.

None of the above is a blocker in my view; 1 and 2 are the two I would fix before this is used in anger.

haraldsteinlechner:
Self-review findings filed: #699 (CSV quoting), #700 (no output file when no image resolves), #701 (CI runs no data-backed tests). Neither #699 nor #700 produces wrong coordinates; both are present in the published 0.2.0.

PR #693, opened 2026-08-19
docs(sbmt): document the SBMT import feature and cover it with tests

haraldsteinlechner:
Documents the SBMT (Small Body Mapping Tool) annotation import, which shipped in 6.0.0 without a `docs/` page, and backs it with tests that run without the external fixtures.

## Documentation

New `docs/SbmtImport.md` covering:

- how to use the import, and why imported groups are collapsed and chunked
- coordinate-frame behaviour — only km → m plus an identity trafo is applied, which is why Dimorphos catalogs land on the body and Didymos ones do not
- the 17-column point and 18-column ellipse/circle layouts, with the indices the parsers actually read
- the ellipse tangent-plane construction, the 60-sample boundary, and the two approximations baked in (radial "up" rather than the mesh normal; the pole fallback)
- why `dnsResults` / `results` are deliberately not computed during import
- known limitations: the dropped reference-frame label, the hardcoded `DIMORPHOS_SHM`, unimplemented line/polyline/polygon, and only the first selected file being imported

## Tests

`SbmtImportAlignmentTest` goes from 4 tests to 25.

The four existing tests all required the external catalogs, and two of them parsed SBMT rows with local helper functions that duplicated the importer rather than calling it — so they passed even if `SbmtImporter` changed. They now go through `SbmtImporter` directly.

Twenty new tests synthesize SBMT files in a temp directory and always run:

- header detection across all six kinds, plus case/whitespace tolerance, unknown kinds, a missing header, and `startImporter` failing loudly on one
- point rows: column mapping, km → m scaling, colour fallback to magenta, quote stripping, trafo applied *after* the unit scale, and rejection of comments, blanks and short rows
- ellipse rows: planarity, centroid at the declared centre, semi-major/semi-minor lengths, `regularAngle` 0 and 90 orientation against an independently computed basis, `flattening = 1` degenerating to a circle, and circle files dispatching through the ellipse parser
- line/polyline/polygon files importing as empty rather than throwing
- `importSbmt` grouping: flat and collapsed below the 100-entry chunk threshold, bucketed above it with no leaves lost, and unique annotation keys

Verified both ways:

- with the fixtures present: **25 passed, 0 failed** (the 4,758-ellipse Dimorphos catalog imports in 0.39 s)
- with `PRO3D_SBMT_TESTDATA` pointed at a nonexistent directory: **20 passed, 5 ignored**

The external fixture root still defaults to `C:\pro3ddata\shapemodels\testdata` — note this is a *different* root from the suite-wide `--testdatasource` (`C:\pro3ddata\testdata`), which the SBMT fixtures were never moved under. It is now overridable via `PRO3D_SBMT_TESTDATA`. Unifying the two roots would mean relocating the data, so it is left as a follow-up.

## Incidental

Six files referenced `plans/sbmtImport.md`, which now lives at `plans/archive/sbmtImport.md` — repointed. The `SbmtImporter` module header still said "v1: points only" after ellipses shipped; refreshed. Added the importer to the `ai/DOMAIN.md` table.

No behaviour changes — documentation, comments and tests only.


haraldsteinlechner:
test's only run on my box, this needs to be reworked.

haraldsteinlechner:
Follow-up from the body ("unifying the two roots ... left as a follow-up") is now done.

Fixtures resolve from a `PRo3D.Resources.TestData` checkout under `imports/`, the same way the rest of the data-backed suite resolves it: `PRO3D_TEST_DATA` first, `--testdatasource` second. `allTests` takes `TestParameters` so the CLI flag reaches the list; the NUnit entry point passes `None` and falls back to the variable.

The committed `imports/basicSBMT-dimorphos-v4` export is a real SBMT v4 dump of Dimorphos, one file per structure kind. The ellipse-planarity test moved off `Didy_Boulders_Paj_Tusb_Lucch` onto its ellipses file, and a new test imports its points file end to end — headers and all — checking km → m against the independent radius column of `centerLLR`.

`pointOnPike.points.txt`, `anno.json` and `Dimo_Bould_Glob_7_Maurizio` are not redistributable and stay outside the checkout. They are searched under `<root>/imports` first, so dropping them in is enough; `PRO3D_SBMT_TESTDATA` still finds them where they are.

| available | result |
|---|---|
| checkout only, via `PRO3D_TEST_DATA` | 22 pass, 4 skip |
| checkout only, via `--testdatasource` | 22 pass, 4 skip |
| checkout + external catalogs | 26 pass, 0 skip |
| neither | 20 pass, 6 skip |

`sbmtimport.circles.txt`, `.paths.xml` and `.polygons.xml` ship in the fixture set but are not exercised yet — circles are covered synthetically, and lines/polygons are still unimplemented in the importer.

`docs/SbmtImport.md` updated to match. Branch merged up to `develop`.

haraldsteinlechner:
Yes on both counts, and now fixed.

**The external catalogs** are Maurizio's `Dimo_Bould_Glob_7_Maurizio` (~4,800 ellipses on Dimorphos, used for the bulk-import and drawing-model timings) and the alignment pair `pointOnPike.points.txt` + `anno.json`. `Didy_Boulders_Paj_Tusb_Lucch` is no longer needed — the ellipse-planarity test moved onto the committed `imports/basicSBMT-dimorphos-v4` export.

**The private path was still there** — `C:\pro3ddata\shapemodels\testdata`, hardcoded as the `PRO3D_SBMT_TESTDATA` default. And in three other places: `OpcSidecarTests` (`PRO3D_BDS_OPC` and `PRO3D_AARA_OPC` defaults, both under `C:\pro3ddata\HERA\OPCUpdate`) and `TriangleSetTests` twice (`C:\pro3ddata\testdata\Dimorphos_DRACO1\…`).

All four now go through `TestUtils.Roots`, which names the two roots the suite actually has:

| root | is | resolved from |
|---|---|---|
| `testData` | a `PRo3D.Resources.TestData` checkout — public, versioned, where a fixture belongs unless it can't be committed | `PRO3D_TEST_DATA`, then `--testdatasource` |
| `privateData` | whatever local pile the rest sits in | `PRO3D_PRIVATE_TESTDATA`, default `C:\pro3ddata` |

The narrow overrides (`PRO3D_SBMT_TESTDATA`, `PRO3D_AARA_OPC`, `PRO3D_BDS_OPC`) still name an individual directory where one sits outside the tree.

`TriangleSetTests` was miscategorised: the Dimorphos patch it reads is checkout data, so it takes `testData` first and only falls back to `<private>/testdata`.

One deliberate choice: setting `PRO3D_PRIVATE_TESTDATA` wins outright. Pointing it somewhere that doesn't exist yields `None` rather than quietly reverting to `C:\pro3ddata` — a typo should show up as skipped tests, not as tests silently reading whatever is at the old path. Leaving it unset still falls back to `C:\pro3ddata`, so existing setups are unaffected.

| list | result |
|---|---|
| `sbmtImport` — both roots | 26 pass, 0 skip |
| `sbmtImport` — checkout only | 22 pass, 4 skip |
| `sbmtImport` — private only | 24 pass, 2 skip |
| `sbmtImport` — neither | 20 pass, 6 skip |
| `sbmtImport` — variable unset (falls back) | 26 pass, 0 skip |
| `TriangleSet` — either root / neither | 14 pass, 0 skip / 12 pass, 2 skip |
| `OpcSidecar` — private root / without | 10 pass, 0 skip / 8 pass, 2 skip |

`docs/SbmtImport.md` and `docs/VertexAttributes.md` updated.

Note this widens the PR past SBMT into `TestUtils`, `OpcSidecarTests` and `TriangleSetTests` — happy to split that into its own PR if you'd rather keep this one to the import docs.

PR #714, opened 2026-08-26
Fix mapview

sophie-sudoku:


haraldsteinlechner (commented):
## Review

The core fix is sound: the `tanh` → `atan` corrections (both the drag angle and the fov in `Navigation.fs`), the per-message radius refresh, the ENU north fix, and leaving `exploreCenter` alone all address the reported freeze, and the history in `docs/MapView.md` matches what the code actually did. The tests pin the right things.

What follows are seven map-view defects that survive or are introduced by this change, plus some quality issues. Two of them are the same class of failure this PR fixes — frozen or runaway navigation — just on different paths. Line numbers are against `bc345416`.

### Real bugs

**1. ENU/JPL frames do have a singularity, and the new code declares they don't.**

`blocksPole` (`MapViewCameraController.fs:129-135`) short-circuits to `false` whenever `polarAxis` is `None`, commented "constant-up frame: no singularity". The *frame* has no positional singularity, but `updateCameraForMapView` still calls `CameraView.lookAt location V3d.OOO frame.north`, and for ENU `north` is the constant `V3d.OIO`. Drag the camera onto the world Y axis (ENU) or X axis (JPL) and `forward` becomes parallel to `north`; Aardvark's `lookAt` computes `cross(forward, sky) |> normalize`, which yields NaN. The camera view is then NaN for every subsequent frame — a dead viewport until the user switches modes. Vertical drag walks straight into it with no clamp anywhere, and MapView is offered for ENU/JPL (`Navigation.fs:265` disables it only for `Planet.None`).

This was latent before as well, but the PR *moves* ENU's singular axis (±X → ±Y) and adds a comment asserting it cannot happen. Fix is small: give ENU/JPL a `polarAxis` of their own north axis instead of `None`.

**2. Right-drag zoom is not height-scaled at all.**

`MapViewCameraController.fs:409` still uses `step = -model.zoomFactor * exp(sensitivity) * forward * delta.Y` with `zoomFactor = 0.01`, never overridden for MapView (`Navigation.fs` sets `sensitivity`, `orbitCenter`, `targetPhiTheta`, `panFactor`, `rotationFactor` — not this one). At 500 km above Mars a full-window right-drag moves the camera roughly 14 m, so the control looks dead.

Both `docs/MapView.md` ("every navigation step is scaled by how high the camera is above the ground", with right-drag listed as zoom) and the `heightAboveSurface` docstring ("Pan speed, zoom step and the zoom-in limit all scale with this") describe behaviour this branch does not implement. Its only limit is `direction > 0`, which stops the camera at the *body centre* — so near the ground it zooms straight through the terrain.

**3. The height floor is 34 m on Mars, not "a small positive value".**

`heightAboveSurface` (`MapViewCameraController.fs:145-148`) returns `max (distanceToCentre - radius) (max (radius * 1e-5) 1.0)`. With `rotationFactor` genuinely holding the body radius (~3.39e6 on Mars, which TC-2.5 asserts), the floor is 33.9 m on Mars and 63.7 m on Earth. A camera 2 m above an outcrop then pans as if it were 34 m up — 17x too fast on Mars, 32x on Earth.

The `1e-5 * radius` choice may well be deliberate as a scale-relative floor, but the doc comment misdescribes it, and the near-ground behaviour is a problem in exactly the close-inspection case PRo3D exists for.

**4. The wheel zoom-in guard is now dead code, and nothing replaced the old floor.**

Since `cam.Forward` is unit, `step.Length = distancetoSurface * |moveVec.Z| * sensitivity * dt`, so the guard `if step.Length > distancetoSurface` (`MapViewCameraController.fs:211`) reduces to `|moveVec.Z| * sensitivity * dt > 1` — the floored height cancels from both sides. With one wheel notch (|Z| = 10), default sensitivity (factor 0.15) and a ~16 ms tick, that product is ≈ 0.024; it would need `dt > 0.67 s` to ever fire.

The old unfloored `distancetoSurface` could go negative and hard-stopped zoom-in below the surface. Removing that was clearly intentional (it caused a freeze), but nothing bounds it now: repeated notches descend a constant ~0.8 m each on Mars, through the terrain, past the reference surface, toward the body centre.

**5. ENU/JPL navigation scale is geometrically wrong, and can freeze.**

For non-planetary frames `mapViewRadius` falls back to `|exploreCenter|` (`Navigation.fs:80-83`), and `heightAboveSurface` then computes `|cameraLocation| - |exploreCenter|`, which is not a height in a flat scene.

With `exploreCenter` at (5000, 5000, 0) and the camera at (5000, 5000, 200), the "height" comes out as 2.8 m instead of 200 — pan roughly 70x too slow. Move the camera to (0, 0, 200) and the subtraction goes negative, floors to 1.0 m, and the drag angle collapses to ~0.0003 rad: controls frozen. That is this PR's own headline symptom, reachable in any ENU/JPL scene.

Compounding it: MapView no longer zeroes `exploreCenter` and never writes it, so a stale ArcBall pick becomes a permanent bogus radius. (Not zeroing it is right — but the fallback then needs a better source of scale.)

**6. Wheel zoom is a no-op on trackpads and Firefox.**

`MapViewCameraController.fs:328`: `moveVec + V3d.OOI * float (int delta.Y) * 10.0`. Aardvark's `onWheelPrevent` divides the browser's `deltaY` by -100 before it reaches the handler, so a standard notch arrives as ∓1.0, but a precision trackpad (deltaY 1–99) arrives as e.g. -0.3 and Firefox line-mode (deltaY 3) as -0.03. `int` truncates all of those to zero. Because `moveVec` is reset every frame (line 275), the fractions can never accumulate into a step either.

Pre-existing line, but it sits in the function this PR is repairing and the fix is to drop the `int`.

**7. The pole guard can be leapfrogged.**

The ±0.999 band is only ~5.1° wide and there is no per-step angle clamp, while `visibleAngle` approaches π at high altitude — so a coalesced fast drag of a few percent of window height can cross the pole without ever landing inside the band. Worse, the dot product is symmetric about the pole, so a step that *does* land inside the band on the far side reads as motion *away* from it and passes. The result is a sudden ~180° roll of the map, not a crash.

The PR preserved the old guard's exact band-plus-monotonicity structure, so this is inherited rather than a regression — noting it because the generalisation to `frame.polarAxis` was the natural moment to fix it.

### Quality

**Native SPICE call on every message.** `rotationFactor = mapViewRadius planet model` (`Navigation.fs:164`) is computed on *every* MapView message, including inert `Down`/`Up`/`Blur`/`Nop`, and `StepTime` is pumped every 10 ms. For the six planetographic bodies each call is a native P/Invoke (`Xyz2LatLonAlt`), where pre-PR it was one `Vec.Length` on mode entry. The bookmark-restore motivation is real, but satisfiable more cheaply — recompute on entry plus when the stored value is still the 0.01 default.

**`blocksPole` rebuilds a frame it doesn't use.** Both call sites (`:266` in the 10 ms tick, `:396` in the drag path) build a full `mapFrame` — including `getUpVector`, itself up to two more native calls — purely to read `polarAxis`, which depends only on `planet` and never on position. A planet-only helper would do.

**Axis conventions now live in three places** — `mapFrame`, `CooTransformation.getUpVector`, and `ReferenceSystem.updateCoordSystemAt` — and have already drifted: for `Planet.None`, `mapFrame` groups with ENU (north = +Y) while the reference system groups it with JPL (north = +X). Latent today because the dropdown disables MapView for `Planet.None`, but it is the same 90°-rotation bug class this PR just fixed for ENU. A single shared source for the ENU/NED axes would collapse it.

**The `StepTime` and `Move` branches duplicate the whole orbit pipeline near-verbatim** — `heightAboveSurface` → half-patch → `visibleAngle` → translate/rotate/translate → `blocksPole` → `CameraView` reconstruction, comment included. The cost is visible in this very PR: both the `tanh` → `atan` correction and the `blocksPole` guard had to be applied twice, once per copy. Only the rotation-axis construction differs (WASD × dt vs drag delta), so one "orbit about centre by (angleX, angleY) with pole guard" helper would serve both.

**Doc/code mismatch on `rotationFactor`.** `docs/MapView.md:121-124` says a camera that has been through MapView "should not be handed to FreeFly without resetting it" because `SetNavigationMode FreeFly` "rebuilds the view for exactly this kind of reason". That branch rebuilds only the view and never resets `rotationFactor`. No consequence today — nothing reads the field in the controllers PRo3D actually uses — but the doc claim is inaccurate as written.

**Silent SPICE fallback.** `mapViewRadius` swallows a failed planetographic lookup and substitutes a heuristic with no log line. Kernel-less Mars is essentially a CI condition, but `switchKernel` and a custom `defaultSpiceKernelPath` make it reachable in production, where degraded pan/zoom sensitivity would have no diagnostic trace. A one-time `Log.warn` would match the codebase habit.

**Minor:** `update Planet.None` in `start()` (`:487`) plus the new `&& model.orbitCenter.IsSome` on the zoom branch makes every interaction except the wheel inert in the standalone test app, since `initial.orbitCenter` is `None` and `Pick` is not wired up. Dead code only — `start()` has no callers.

PR #734, opened 2026-09-10
Run the HERA SPICE-kernel tests on CI, and fix what that uncovered

haraldsteinlechner:
The kernel-backed tests only ever ran on a workstation with a full ESA mirror in a sibling `spice` directory, so CI covered the `--skip-hera` subset and the projection fixtures went unchecked on every PR. This gets the kernels onto GitHub Actions — and fixes the two SPICE bugs that surfaced the moment the whole suite ran with kernels present.

## Getting the kernels (`ci: run the HERA SPICE-kernel tests on GitHub Actions`)

Fetch what the suite actually loads instead of mirroring anything. A HERA meta-kernel names its kernels in `KERNELS_TO_LOAD`, so the closure of the four meta-kernels the tests open is **119 files, 1.23 GiB**, against ~11 GB for the dataset (and a git repo carrying every version ever published). ESA serves it over public HTTPS with range support, so `scripts/fetch-spice-kernels.sh` is resumable and runs 8 downloads at a time.

**Pinned by version.** The unversioned names the tests open — `hera_ops.tm`, `hera_plan.tm` — are copies of whatever release is newest, and ESA publishes every few weeks, moving the previous file into `former_versions/` as it goes. It did exactly that to the release this branch was written against while the branch was being written. `scripts/spice-kernels.pins` names versioned files, resolved in `mk/` then `mk/former_versions/`, and the aliased ones are written into `mk/` under the plain names — rewriting `PATH_VALUES` back to `'..'`, the single line archiving changes. The result is byte-identical to a workstation mirror's `hera_ops.tm`.

**Safe to run on a cache hit.** The script verifies the tree against its own manifest offline and exits in about a second; it repairs a partial restore rather than letting it surface later as a puzzling error inside CSPICE. The manifest records which pins produced it, so a tree restored via `restore-keys` after a pin bump refetches the changed closure instead of being mistaken for complete.

Tests find the tree through `PRO3D_SPICE_KERNELS` (the variable `pro3d-tool` already reads, equally tolerant of root-or-`kernels`), falling back to the sibling mirror, so existing workstations are unaffected. The job is Linux-only: kernels are platform-independent data, and a 1.3 GB cache entry per OS would push the paket caches out of the repository's 10 GB budget.

## What that uncovered (`fix(tests): stop the kernel swap from corrupting SPICE state`)

Running the whole suite with kernels present died mid-run — `SPICE(BADSUBSCRIPT)` in `dafah`, then every later lookup failing with `SPICE(DAFNOSUCHHANDLE)`, then the process going down — and on the way there the two `pro3d-tool` tests reported *"the loaded kernel has no coverage at this epoch"* for epochs their kernel covers perfectly well when the test runs on its own. Two causes, both about how the kernel gets loaded rather than about the kernel.

**The working directory.** SPICE stores the file names a meta-kernel produces exactly as they read at furnsh time, and DAF re-opens binary kernels lazily, by the stored name, long afterwards. ESA's meta-kernels name them relatively (`PATH_VALUES` is `'..'` in `mk/`, `'../..'` in `mk/former_versions/`), so `ensureKernelAt` chdir'd into the meta-kernel's directory and parked the process there. That holds only while every meta-kernel lives in the same directory; once fixtures started naming archived kernels, a swap moved the working directory out from under names DAF had already stored. `PRo3D.Base.CooTransformation` already solved this for the viewer — `materializeMetaKernel` rewrites `PATH_VALUES` to an absolute directory before the furnsh — so the tests now go through `switchKernel` and inherit it. Nothing chdirs any more.

**The other thread.** Swapping is `DeInit + Init + furnsh`: it empties the kernel pool. The concurrency test in `InstrumentProjectionComparisonTest` ran 25 HSH computations against 25 ASPECT ones, and those two fixtures name different meta-kernels — so half the threads were freeing the kernels the other half were reading. `InstrumentProjection` now exposes `withSpiceLock` so a caller replacing the pool takes the same lock the readers do, and the concurrent section uses one fixture throughout. That is not a loss of coverage: with two kernels, an HSH lookup answered by the plan kernel returned `None` and was dropped by the `Array.choose`, so the drift check was measuring whatever happened to survive. Now every result must resolve and match the baseline.

## Verified

- All 119 closure URLs checked against ESA: sizes match a workstation mirror byte for byte.
- Fetch script exercised end to end — cold, warm, missing file, truncated, oversized, pin bump, malformed meta-kernel — plus real HTTPS download and resume-repair.
- Full suite with kernels and private data: **482 tests, 480 passed, 13 ignored, 2 failed** (it used to crash mid-run). Zero `BADSUBSCRIPT` / `DAFNOSUCHHANDLE` in the log.
- CI-shaped run (kernels, no private data, no GL): **472 passed, 21 ignored, 2 failed**.
- Kernel-free subset (`runTests.sh`, what the four-OS matrix runs): **445 passed**.

## Known failure this does not fix

Both remaining failures are `snapshot-sun-lighting`. Its subprocess fails to render its *first* frame (`Could not render image …frame_a.png`, `col = None` in `RenderingUtils.fs:238`) while `frame_b` renders and saves fine — fresh process, one kernel, no swaps, no SPICE error in its log; the preceding line is `[Viewer] Message not handled: SetRenderViewportSize [640, 480]`. There is also an unrelated unhandled `Task.Dispose` exception at `PRo3D.Snapshots/Program.fs:234` on shutdown. Pre-existing, in `PRo3D.SimulatedViews`, and it reproduces in isolation.

Worth knowing: these tests gate on `PRO3D_SPICE_KERNELS`, so they have been skipping on machines that never set it. Now that setting it is the documented way to run the suite, anyone with the `C:\pro3ddata` workshop fixture and a GPU will start seeing them fail. On CI they still skip — no fixture, no GL context, no `PRo3D.Snapshots.exe`.

## Docs

`docs/tests/SpiceKernels.md` is new — where the kernels come from, why the pins exist and how to bump them, what CI does, and the rule the second commit is about: load kernels through `switchKernel` under `withSpiceLock`, never by chdir'ing. Linked from `docs/spice.md` and `docs/tests/TestData.md`; `AGENTS.md` and the `runAllTests` headers point at it too.

https://claude.ai/code/session_01WcCDi43HPticvCkzcQtZUo

### DIFF
--- a/docs/tests/SpiceKernels.md
+++ b/docs/tests/SpiceKernels.md
@@ -1,20 +1,17 @@
 # SPICE kernels for the test suite
 
-Part of the test suite drives real SPICE geometry: HERA spacecraft states, instrument
-frames, the Didymos-system body frames the mbi fixtures in `src/Tests/data` were
-generated against. That needs ESA's HERA mission kernels, which are public but far too
-large to commit or to carry in a submodule.
-
-[`scripts/fetch-spice-kernels.sh`](../../scripts/fetch-spice-kernels.sh) downloads
-exactly the kernels the suite loads — **~120 files, ~1.3 GB** — and CI caches them.
+Part of the suite drives real SPICE geometry — HERA spacecraft states, instrument frames,
+the Didymos-system frames the mbi fixtures in `src/Tests/data` were generated against.
+That needs ESA's HERA mission kernels: public, but too large to commit.
 
 | | |
 |---|---|
 | Source | `https://spiftp.esac.esa.int/data/SPICE/HERA/kernels/` (public, no credentials) |
 | Full dataset | ~11 GB — do **not** mirror it |
-| What the suite needs | ~1.3 GB, the closure of four meta-kernels |
+| What the suite needs | 119 files, 1.23 GiB — the closure of four meta-kernels |
+| Fetched by | [`scripts/fetch-spice-kernels.sh`](../../scripts/fetch-spice-kernels.sh) |
 | Pinned in | [`scripts/spice-kernels.pins`](../../scripts/spice-kernels.pins) |
-| Read by | `HeraSpiceTests.kernelsDir`, via `$PRO3D_SPICE_KERNELS` |
+| Found via | `$PRO3D_SPICE_KERNELS`, read by `HeraSpiceTests.kernelsDir` |
 
 ## Getting them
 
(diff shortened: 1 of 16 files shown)


src/Tests/InstrumentProjectionComparisonTest.fs
```diff
@@ -197,24 +197,39 @@ let tests () =
                 Expect.isSome hit "quaternion-based center ray should hit round Didymos"
         }
 
-        // Reproduces the original race (HSH + ASPECT interleaved across threads) and checks
-        // every concurrent HSH result still matches a fresh sequential baseline.
+        // Reproduces the original race -- projectOntoQuat and projectOnto, which are each
+        // several native calls, running interleaved across threads -- and checks every
+        // concurrent result still matches a fresh sequential baseline.
+        //
+        // All 50 items compute the HSH projection, and deliberately so: they used to be
```
haraldsteinlechner: insane comment

docs/tests/SpiceKernels.md
haraldsteinlechner, line 1: tighten!

scripts/fetch-spice-kernels.sh
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/env bash
```
haraldsteinlechner, line 1: tighten comment

src/PRo3D.Base/InstrumentProjection.fs
```diff
@@ -95,6 +95,17 @@ module InstrumentProjection =
     // per-thread reentrant, so nesting (projectOnto* calls getLookAt*) is safe.
     let private spiceCallLock = obj()
 
+    /// Runs `f` as one unit against every SPICE-touching call in this module.
```
haraldsteinlechner: shorten, fact based

src/Tests/HeraSpiceTests.fs
```diff
@@ -15,18 +15,46 @@ open Aardvark.Base
...
-
-// HERA tests need the (non-public) HERA mission kernels. They self-skip when
-// those are absent (e.g. in CI), or when --skip-hera is passed (runTests uses
-// this for a deterministic kernel-free run). Kernel-independent SPICE coverage
-// lives in SpiceTests.fs and always runs. (Adopted from releases/6.0.0.)
+
+/// The `kernels` directory of a HERA SPICE dataset, i.e. the directory holding
+/// `mk`, `ck`, `spk`, ... .
+///
+/// PRO3D_SPICE_KERNELS is the documented way in - the same variable PRo3D.Tool reads,
```
haraldsteinlechner: condense

src/Tests/HeraSpiceTests.fs
```diff
@@ -52,14 +89,39 @@ let init () =
...
 // nothing to unload a single kernel). So: track which single kernel is active,
-// and swap via DeInit+Init+AddSpiceKernel whenever a scenario needs a different
-// one -- verified empirically that a fresh Init with just the target kernel
-// resolves correctly (both for an isolated hera_ops_v172 and an isolated
-// hera_plan). Every caller (fixture-driven or generic) must re-request its
-// kernel immediately before making SPICE calls, since anyone could have swapped
-// it since the caller last checked.
-let private spiceLock = obj()
-let private activeKernel : string option ref = ref None
+// and reset the pool whenever a scenario needs a different one -- verified
```
haraldsteinlechner: shorten, fact based, add to conceptual part to relevant docs

src/Tests/InstrumentProjectionComparisonTest.fs
haraldsteinlechner: Cut to ~10 lines in ad66e5f6: what the race is, and why one fixture rather than two (different meta-kernels, only one loads at a time, so half the threads were freeing what the others read). The rest is gone.

docs/tests/SpiceKernels.md
haraldsteinlechner: Rewritten in ad66e5f6 — roughly half the length. Kept the tables, the pins, the bump procedure and the CI mechanics; dropped the padding. It also gained the conceptual part that used to sit in HeraSpiceTests.fs (why absolute PATH_VALUES, why the lock).

scripts/fetch-spice-kernels.sh
haraldsteinlechner: Tightened throughout in ad66e5f6, header and inline (97 lines of comment down to 36 changed). Each one now states the non-obvious fact and stops; the reasoning is in docs/tests/SpiceKernels.md.

src/PRo3D.Base/InstrumentProjection.fs
haraldsteinlechner: Down to four lines in ad66e5f6: what it is for (callers that replace the pool), the two concrete failure modes (a swap between the two calls of one projection answers it from two kernel sets; a swap inside a native call is an access violation), and that it is reentrant.

src/Tests/HeraSpiceTests.fs
haraldsteinlechner: Condensed to four lines in ad66e5f6 — where the directory comes from, the fallback, and that not-found means the tests skip. Details moved to docs/tests/SpiceKernels.md.

src/Tests/HeraSpiceTests.fs
haraldsteinlechner: Shortened to two short paragraphs in ad66e5f6: one meta-kernel at a time and why, then the two rules for switching (absolute PATH_VALUES, hold the projection lock) with a pointer. The DAF/working-directory explanation now lives in docs/tests/SpiceKernels.md under "Loading a kernel".

Issue #741, opened 2026-09-10
Multi-image projection: follow-ups left out of #735


haraldsteinlechner:
What #735 (multi-image projection fixes + validation, landing together with #720 into `releases/6.2.0`) deliberately left out. Each item states what is the case in the code as merged, and how we know.

## Correctness and design

1. **Surface transformations do not reach the projection.** The per-patch projector matrices are `projector * Local2Global` (`ImageProjectionOpcRendering.fs`, `projectionUniformMap`). The surface model trafo — SPICE placement *and* the user Transformation (translation, yaw/pitch/roll, scaling, pivot) plus `preTransform` — is left out, because including it applied the body rotation twice (correlation 0.028 → 0.976 when removed). Consequence: a registration correction made by transforming a surface is not seen by the projection. Documented as a known limitation in `docs/MultiImageProjection.md`. Fix: compose the user part of the transformation, but not the SPICE part, into the projector matrices.

2. **The projector-facing test is not a visibility test.** `stableImageProjectionStack` accepts a fragment when `ProjectedStackTrafos[i].TransformDir(normal).Z < 0`, i.e. the view-space normal's z. There is no occlusion test, so terrain the projector cannot see because other terrain is in front (ridges, boulders) is painted too. It also compares against the projector's axis, not the ray to the fragment, so faces near the image edge are misclassified by up to half the field of view (≈2.8° for AFC-1, more for wider instruments). A depth map rendered from each projector would be exact, handle occlusion and make the winding vote below unnecessary.

3. **`NormalFlip` depends on a heuristic, and its flip path has no test.** `NormalWinding.estimate` votes over ~100 faces of each hierarchy's root patch whether they point away from the body-fixed origin. That assumes a roughly star-shaped body seen at orbital scale; it is not reliable for near-vertical terrain (rover outcrops), where the dot product with the radial direction is near 0. The only inward-wound dataset (`Dimorphos_0_Meridian`) was retired; `NormalFlip 1` was validated once with it (r 0.9900 at zero shift, 2026-09-10) and is not covered by any test now.

[...]

PR #753, opened 2026-09-14
imageTrafo: read the hoisted surface frame instead of building a map per evaluation

haraldsteinlechner:
Small follow-up from #741 (item 12), kept separate from #750 as agreed.

`ProjectedImagesListHelpers.getProjectedImageData`'s `imageTrafo` built a fresh `AVal.map` over `surfaceReferenceSystem` **inside** its `AVal.custom` on every evaluation, and read it through the token. The stack code in the same function already used a hoisted `surfaceReferenceFrameA` for exactly this. `imageTrafo` now shares it, and the definition moves up so both can see it.

## Why this is a correctness fix

In FSharp.Data.Adaptive, dependency edges point forward and are **weak** (`Outputs : IWeakOutputSet`). A consumer keeps its inputs alive only through references it holds itself:
- `map`, `bind` and `adaptive {}` keep their input or inner node in a field.
- `AVal.custom` keeps only its compute function.

A node created during a custom's evaluation is therefore owned by nothing. Once the GC collects it, the chain *input → node → custom* is gone. The custom is never marked again and keeps serving its cached value, without even being out of date.

Measured with FSharp.Data.Adaptive 1.2.26, three runs each: evaluate, `GC.Collect()`, change the input, evaluate again.

| pattern | after the change |
|---|---|
| fresh `AVal.map` inside `AVal.custom`, read with the token (the old `imageTrafo`) | **stale**, and not out of date |
| hoisted map | follows |
| fresh map inside `AVal.bind`'s lambda | follows (bind holds its inner) |
| fresh map under `let!` in `adaptive {}` | follows |

Without a GC in between, the old code works, which is what my first commit message wrongly concluded from. That message is corrected by the second commit, so squashing on merge keeps the history accurate. The pattern and the mechanism were already pinned by `src/Tests/AdaptiveNestingTests.fs` from the PackedRendering fix.

**Impact here is small:** `imageTrafo` only feeds the legacy single-image `ProjectedImageModelViewProj` uniform, which the viewer's surface effect doesn't sample. `SunShadowMap.sceneBounds` has the same bug with a visible effect and is not addressed here: it builds `surfacePlacement` inside its custom, so after a GC the shadow camera's bounds stop following surface transformations and time changes.

## Tests

Expecto 491 passed, 23 ignored, 0 failed.


src/PRo3D.GIS/ProjectedImagesListHelpers.fs
```diff
@@ -105,17 +105,24 @@ module ProjectedImagesListAppHelper =
                 Trafo3d.RotationXInDegrees(b.yaw.value) * Trafo3d.RotationYInDegrees(b.pitch.value) * Trafo3d.RotationZInDegrees(b.roll.value)
             )
         let boresightAdjustment = computeBoresight g.projectedImageList.boresightAdjustment
-        let imageTrafo = 
-            AVal.custom (fun t -> 
+        // Created once, outside the evaluations, and shared by imageTrafo and the
```
haraldsteinlechner: long story, shorten

haraldsteinlechner: Shortened to two lines in 1a5474df; the full explanation stays in the PR description and AdaptiveNestingTests.fs.

PR #759, opened 2026-09-15
Scene body: global planet and GIS observation are one setting (#758)

### LINKED ISSUE #758: Single scene body: global planet and GIS observation share one source (MapView + image projection for single-body scenes)
## Problem

PRo3D has two independent "scene reference" settings that disagree:

- **Global planet** (`ReferenceSystem.planet`) — assumes world = planet centre, planet body-fixed axes. Drives up/north, lat/lon, bearing/dip, MapView (pole = +Z, body radius), sky projection, scale bars.
- **GIS observation** (observed body + reference frame, per-surface body/frame) — places surfaces via `transformBody(surfaceBody, surfaceFrame -> observer, observerFrame)`.

Consequences:

- A scene set up only in the GIS view has `planet = None`, so MapView is greyed out (`Navigation.UI.viewNavigationModes`). First import's `InferCoordSystem` also overwrites a manually chosen small body (Dimorphos) with `None`.
- When the GIS frame is not the body's fixed frame (e.g. the HERA test scene: observed body Dimorphos, frame J2000), world is no longer body-fixed, so every planet-based computation uses the wrong pole/axes; points from both spaces (untransformed `globalBB` vs. world picks) are mixed into the reference system, which shows up as a broken up vector.
- With a non-identity SPICE trafo there are genuine double/mixed applications: annotation control points are stored in world space but re-transformed by the non-packed renderer (Snapshots); segment samples go through `toLocal` while control points do not; the image projection composition is suspect.

All of this disappears when the SPICE trafo of the scene body's own surfaces is the identity, i.e. observed body = surface body and frame = that body's fixed frame.

## Short-term scope (this issue)

Make single-body scenes consistent so image projection and MapView work without extra setup:

- One **scene body**: the global planet dropdown and the GIS "observed body" edit the same value (single write point keeps `referenceSystem.planet` and the GIS observation in sync). Planet <-> SPICE body + fixed frame table (Didymos -> DIDYMOS_FIXED, Dimorphos -> DIMORPHOS_FIXED, Mars -> IAU_MARS, ...).
- For new setups the scene frame is the body's fixed frame; the GIS reference-frame field is kept as the future scene-frame field (no format change).
- Surfaces without an explicit GIS assignment inherit the scene body.
- `transformBody` short-circuits to identity for body == observer in its own frame (no SPICE kernel needed, no log spam for plain Mars scenes).
- First import no longer overwrites a scene body.
- Sequenced bookmarks contribute observation time + camera source only; their observer/frame no longer switch the scene.
- Old scenes whose GIS frame is not body-fixed load unchanged (legacy behaviour).
- Remove hardcoded `MARS` / `IAU_MARS` fallbacks in GIS code paths.

## Deferred (must stay possible)

A freely chosen scene frame (e.g. J2000): one time-dependent body-fixed -> world transform, authored geometry stored body-fixed, planet features routed through that transform, and fixing the annotation / pivot / projection composition issues above.

haraldsteinlechner:
> **This PR is half of the work.** It makes single-body scenes consistent (the scene body's own surfaces get the identity placement, so nothing can be transformed twice). Anything with a *non-identity* SPICE placement — multi-body scenes, scenes still observed in J2000, a spacecraft as observed body — is **still unsafe** and is tracked in #761 (multi-body projection origin: #760). Target: `releases/6.2.0`, so single-body HERA work (projection + map view) is usable now.

Closes #758 (short-term scope). Remaining half: #761. Related: #760.

## What

The global planet (`ReferenceSystem.planet`) and the GIS observation (observed body + reference frame) are now **one setting, the scene body**. Pick *Dimorphos* in either place, load a kernel, add an image: projection, sun lighting, MapView, lat/lon and measurements work without per-surface setup.

## Why

Planet-based code (up/north, lat/lon, bearing/dip, MapView's +Z pole and radius) reads world coordinates as the planet's *body-fixed* coordinates. The GIS places surfaces via `transformBody(surface body/frame -> observed body, observation frame)`. The two agree exactly when the body is observed in its own fixed frame — then the scene body's surfaces get the identity placement. As two independent settings they disagreed:

- GIS-only setups kept `planet = None` → MapView greyed out (and first-import inference reset a chosen Dimorphos to `None`).
- Scenes observed in `J2000` (e.g. `ProjectionTest.pro3d`) were read as body-fixed → broken up vector, wrong MapView pole, rotated lat/lon.

## Changes

- `PRo3D.Base.Gis.SceneBody`: planet ↔ SPICE body + fixed frame (Didymos → `DIDYMOS_FIXED`, Dimorphos → `DIMORPHOS_FIXED`, …), body-fixed test; SPICE names compared case-insensitively.
- `transformBody` short-circuits *body seen from itself in its own frame* to the identity, without SPICE.
- `SceneBodySync` (viewer) is the only writer of planet + observation: planet dropdown → GIS observation; GIS observed body → frame snaps to the fixed frame, planet follows; scene load with a body-fixed observation fills in the planet (new load step before the scale bars, in all four load pipelines).
- Unassigned surfaces inherit the scene body (only while body-fixed); the shadow-map sun falls back to such a surface.
- Bookmarks (interactive + `PRo3D.Snapshots`) contribute **time + camera source only**; observer/frame are ignored. Interactive playback now also applies the time for bookmarks without a camera source (it applied nothing before), matching Snapshots.
- First import doesn't second-guess a body the GIS already observes; `suggestedSystem` keeps bodies it can't recognise.
- Lat/lon on a GIS body uses the body's convention (`tryGetLatLonAltOfBody`; Dimorphos was going through PGRREC by name); sun direction no longer falls back to Mars/`IAU_MARS` and isn't queried while lighting is Off.
- GIS view: frame row shows the fixed frame; for a scene in another frame it explains and offers *Use DIMORPHOS_FIXED*; surface dropdowns show *Scene body (…)*; bookmark observation settings hide observer/frame.

## Compatibility

- No model or format change, no Adaptify regeneration.
- Scenes saved with an observation in another frame (J2000) **load unchanged** (planet stays as saved, unassigned surfaces stay unplaced). Switching via the button does not convert saved cameras/bookmarks/annotations.
- Picking a planet now also sets the GIS observation (e.g. Mars → `Mars`/`IAU_MARS`); rendering is unchanged since the placement is the identity.

## Not in this PR (tracked in #761)

Freely chosen scene frame (J2000): one time-dependent body-fixed → world transform for planet features and stored geometry, plus the annotation SPICE-trafo inconsistency (control points world vs. sampled points `toLocal`; non-packed renderer re-applies the trafo), which only shows with a non-identity placement.

## Testing

- Expecto: new `Section12_GisView` TC-12.4 (8 tests: table round-trip, body-fixed detection, identity shortcut without kernel, frame snap, planet ↔ observation mirroring, surface inheritance incl. J2000 legacy, inference, Dimorphos lat/lon). Full kernel-independent suite (`--skip-hera`): 547 passed, 13 ignored, 0 failed.
- Playwright (`tests-ui`, real viewer, HERA kernels), all green:
  - new `scene-body.spec.ts` (3): GIS-only scene gets its planet on load, map view switches and a drag orbits the body (north-up, altitude constant); J2000 scene loads unchanged (planet None, map view off, orange note), *Use DIMORPHOS_FIXED* makes it body-fixed and enables map view; picking Dimorphos in the top bar sets observed body + `DIMORPHOS_FIXED` and the *Scene body (Dimorphos)* surface entry.
  - `projection-e2e`: two new scene-body cases (no surface binding, body-fixed world, planet filled in on load; same epoch and across epochs) reproduce the generated frame exactly like the legacy cases: correlation 0.9948 at zero shift, identity beats mirrors by 0.74, coverage 98.7 %, spill 0.03 %.
  - regression: `hover-flyto`, `looking-at-dimorphos`, `projection-smoke`, `stack-ui`, legacy `projection-e2e` cases pass.
- The UI run found one gap, fixed in 13254d04: fly-to-image only looked at explicit surface bindings.

## Review

An independent review (separate agent) found no blockers; its should-fix items are addressed in 2c327c48: bookmark time animation without a camera source (null-record crash), camera source = observed body, None/ENU/JPL no longer wipe non-body-fixed observations, *Observe \<planet\> as scene body* offer in the GIS panel for planet-set/GIS-empty scenes, shadow-map sun source, half-assignment placeholders, `LoadSerializedScene` reconcile, docs, and viewer-level Expecto tests (TC-12.5, TC-8.5).

Latest runs: Expecto `--skip-hera` 557 passed / 0 failed; Playwright 11/11 (after a fresh Release build). One earlier Playwright run had `hover-flyto` stuck on the loading splash after hovering (not reproducible since: 1 full suite + 10 isolated runs green); per-launch logs are now kept (`tests-ui/artifacts/logs/`) to catch it next time.


haraldsteinlechner:
**Landing this now, knowingly incomplete, because the workshop needs it ASAP.** The findings of the second review below are **not fixed** in this PR; they are deferred to #761 together with the rest of the unfinished half.

Second review (independent agent): no blockers, "OK to merge into `releases/6.2.0`". Open should-fix items:

1. **PRo3D.Snapshots batch files lose their per-bookmark observer/frame** (`SnapshotGenerator.fs`). A scene with GIS bindings + lighting but *no* scene-level observer, whose batch bookmarks carried `DIMORPHOS`/`DIMORPHOS_FIXED`, used to be placed and lit via the bookmarks; now it has no observer → no placement, no projection, no sun, and the run still exits 0. Only a `Log.line` says so. Workaround: set the observed body in the scene (`gisApp.defaultObservationInfo.observer/referenceFrame`). Intended fix: adopt a bookmark's observer/frame only when the scene observes nothing, `Log.warn` when they differ.
2. **Annotations drawn on unbound surfaces persist the *inherited* scene body as their `referenceSystem`** (`Viewer.fs` `matchPickingInteraction` → `AddPointAdv`). On the base branch they persisted `None`. Changing the observed body afterwards displaces them in the non-packed (Snapshots) renderer, and it creates persisted data #761 will have to migrate. It starts as soon as the planet dropdown is touched (a plain Mars scene then persists Mars/`IAU_MARS`). Intended fix: persist only explicit GIS bindings.
3. nit: `GisApp.viewSurfacesInGroups` builds its own scene-body aval instead of `sceneBodyAdaptive`.

**Release notes** are not edited here (a push touching `PRODUCT_RELEASE_NOTES.md` triggers the deploy workflow). Suggested entries for the next 6.2.0 prerelease:

- GIS: **the planet and the GIS observed body are one setting (scene body)** — picking either sets the other; a known body is observed in its fixed frame; surfaces without a GIS binding belong to the scene body, so image projection, sun lighting and MapView work without per-surface setup. Scenes observed in another frame (J2000) load unchanged and offer *Use \<BODY\>_FIXED*. Multi-body scenes are not covered yet (#761)
- GIS: sequenced bookmarks (interactive and `PRo3D.Snapshots`) contribute **observation time and camera source only**; their observer/frame are ignored. Interactive playback now also applies the bookmark time without a camera source
- GIS: sun lighting no longer falls back to Mars/`IAU_MARS` when no observed body is set

Known unexplained, not reproducible since: one Playwright run got stuck on the loading splash after hovering a projected image (`hover-flyto`); per-launch logs are now kept in `tests-ui/artifacts/logs/` to catch it.

PR #762, opened 2026-09-16
OPC surface effect: geometry stage and cross-section discard only when needed (#719)

### LINKED ISSUE #719: Geometry-shader stage dominates OPC frame time even when disabled (13.9x on Apple Silicon; other platforms unmeasured)
The geometry-shader stage in `ViewerUtils.surfaceEffect` dominates OPC frame time, even
when the filter it implements is switched off.

Measured with the OPC screenshot harness (branch `bugs/apple-silicon-crosssection-clip`,
`890cf57a`, plus an unmerged `--benchmark`/`--drop` mode). Dataset `victoria`, offscreen,
median of 5 x 60 frames, Apple M1, OpenGL 4.1.

## Cleanest isolation — only `triangleSizeFilter` differs

| config | ms/frame | fps |
|---|---|---|
| no geometry shader | **6.86** | 146 |
| + `triangleSizeFilter`, **disabled by uniform** | **81.98** | 12 |
| same, stage dropped (sanity check) | 6.70 | 149 |

**+75 ms/frame for a geometry shader that is switched off.** Resolution-independent
(+71.8 ms at 320x200, +67.5 ms at 1920x1200), so it is geometry-stage cost, not fragment work.

## On the full effect stack

| config | ms/frame |
|---|---|
| default — what the viewer composes today | 64.12 |
| triangle filter actually **enabled** | 61.40 |
| drop `triangleSizeFilter` (`generateNormal` remains) | 61.90 |
| drop **both** geometry shaders | **4.61** |

`--dump-glsl` shows the mechanism: the stack emits **one** `#ifdef Geometry` block. FShade
**merges** `triangleSizeFilter` and `generateNormal` into a single GLSL stage, so dropping
either alone leaves the stage — and its cost — in place. Dropping both is **13.9x**.

## Two counter-intuitive results

- **Enabling the filter is faster than leaving it composed-but-disabled** (61.40 vs 64.12):
  it emits fewer triangles. The filtering work is free; the *stage* is the entire cost.
- **Dropping `generateNormal` alone measured 146.21 ms** — 2.3x *slower* than composing
  both. Removing a shader made it slower. Unexplained; likely varying count in the
  merged-vs-single GS. It means **stage cost is not additive**, so any attribution may only
  claim "dropping X saves Y", never "X costs Y".

## What was NOT established

Stated plainly, because the numbers are strong enough to be over-read:

- **Whether this is Mac-specific is unmeasured.** No Windows or Linux comparison was run.
  "Geometry shaders are slow on Apple's GL-over-Metal" is *consistent* with this data but
  not demonstrated by it — the cost may well be present everywhere.
- **The viewer itself was never benchmarked.** All measurements are the harness. That
  `ViewerUtils.surfaceEffect` pays the same cost is an inference from code reading (it
  composes both stages unconditionally), not a measurement. Verify before acting.
- One machine, one dataset, one camera, via a machine-specific data root. Unreplicated.
- The originally reported symptom — the triangle filter feeling slower interactively — **did
  not reproduce offscreen**; it measured slightly faster. The cost was present in both arms
  of that comparison.

## Possible directions

Not chosen — `ai/CONVENTIONS.md` asks that shader-vs-CPU trade-offs be discussed rather than
picked silently:

1. **Conditional composition** — omit `triangleSizeFilter` when a surface has
   `filterByTriangleSize = false` (the default, `SurfaceApp.mk`). Worth only ~3% on its own,
   since `generateNormal` keeps the merged stage alive.
2. **Vertex-stage degenerate emission** instead of a geometry shader.
3. **CPU-side filtering at patch build time.**

Only (2) or (3) reach the 13.9x, and both are larger changes.

## Next step

Measure Windows/Linux with the same harness before treating this as a platform issue, and
benchmark the viewer itself before treating it as a viewer issue.

haraldsteinlechner:
Closes #719 when the macOS numbers confirm it.

The geometry-shader stage (`triangleSizeFilter` + `generateNormal`, merged by FShade into one) dominated OPC frame time on Apple Silicon **even switched off** (#719: 64 → 4.6 ms/frame without it). The cross-section `discard` defeats hidden-surface removal on tile-based GPUs whether or not it discards.

The surface effect now comes in four variants; each surface picks one through `ViewerUtils.surfaceEffectPool` (an Aardvark `Surface.Dynamic`), so switching swaps the GL program without rebuilding render objects or reloading patches.

- **geometry stage**: on while the triangle / distance filters are on, or while a shader reading its face normal can run (projection stack, hover outline, sun shading, shadows). Otherwise `noFaceNormal` writes a constant `LocalNormal`, so it never becomes a vertex attribute.
- **cross-section clip**: on while clipping is enabled and a cross-section exists.

## Startup cost, and the notice you asked for

All variants share one FShade input layout, and building it forces a **full link of every variant** (tens of seconds each for this stack). So:
- `SharedEffectPool` links once for all surfaces and **stores the layout on disk** next to the GL shader cache. A **warm start links nothing** (measured: 269 s → 16 s for a PRo3D.Snapshots run).
- A **cold start** (first start after a shader change) links all four in parallel: **150 s measured here**, against ~45–65 s for the single effect before. It also **compiles every variant up front** on the patch loader thread, so no variant ever compiles on the render thread.
- While that happens the render view shows *"Preparing surface shaders…"*, where "Importing OPCs…" appears. Until now the surfaces were simply absent with no log output (`ai/TESTING.md` already warned about this).

## Testing

New offscreen harness (`SurfaceEffectHarness`): renders a **viewer model's** surfaces through the viewer's own `createGroupedSgs`, so a test loads a scene, renders, changes the model, renders again. Needs a GPU and `PRO3D_TEST_DATA`; skips otherwise.

- `SurfaceEffectSwitchTest`: a filter switch at runtime reaches the OPC patches and draws the same pixels; **a projection added at runtime looks exactly as with the geometry stage from the start**, with sync and async loading and with winding correction off and on.
- `SurfaceEffectVariantTest`: each variant generates GLSL, both switches reach it, no variant needs a vertex input the full effect does not, the cached layout round-trips.
- `tests-ui`: specs now wait for `surfaceShadersReady` before any screenshot, so a cold compile cannot pollute them. `hover-flyto` and `projection-smoke` pass with the same pixel diffs as unchanged 6.2 (0.85 % / 2.47 %).

## What to check on macOS

Frame time on an OPC scene with no projection, no lighting and no filters (the lean variant) against 6.2.0-prerelease002. That is the case #719 measured as 13.9×.

## A variant can fail to link

A cold link of the lean variant failed once here ("[GL] Shader compiler returned errors: failed to link program") and succeeded on the next run; the generated GLSL for this stack is ~22k lines. Precompiling made that a hard failure: an exception out of the first patch prepare, so no surfaces at all. `precompile` now records per variant whether its program compiled and the pool falls back to one that did (the full variant if possible), so the scene renders without that variant's saving rather than not at all. **Worth watching on macOS**, where the driver differs.

## Honest state

- **Not measured on macOS**, and #719's numbers come from a benchmark harness, not the viewer.
- During development, switching the geometry stage in for a projection **did not show the projection** in `tests-ui`. It passes now, in the harness and in `tests-ui`, but I could not explain what fixed it between those builds. The harness test above is there to catch it if it returns.
- Cross-section clip switching has no test.
- Verified after the final edits: Windows build clean, harness tests 5/5, variant tests 5/5, `hover-flyto` and `projection-smoke` pass.
- The intermittent link failure above is not explained, only contained.


haraldsteinlechner:
## The geometry shader may not have been the main cost

Pushed `24c87af8`, which is independent of the variant switching and may matter more on macOS.

**Why the GLSL was 22k lines.** FShade inlines the rest of an effect once per *return path* of a stage, so the generated code is the **product** of the return counts down the stack, not their sum ([krauthaufen/FShade#39](https://github.com/krauthaufen/FShade/issues/39)). Nine stages in the surface effect had two or three returns; the last stage ended up inlined **861 times**.

**The fix** is mechanical — one `return` and a mutable accumulator per fragment stage:

```fsharp
fragment {
    let mutable c = v.c
    if uniform.SomethingEnabled then c <- ...
    return c
}
```

Applied to `markPatchBorders`, `terrainSunShadow`, `hoveredProjectionOutline`, `projectedStackCoverage`, `solarShadingLS`, `mapRadiometry`, `LoDColor`, `secondaryTexture`, `outcropTrace`. No behaviour change.

| | before | after |
|---|---|---|
| GLSL, lean variant | 21971 lines | **830** |
| GLSL, full variant | — | 1016 |
| FShade link, all 4 variants, cold | minutes | **8.3 s** |
| GL compile, all 4 variants | — | 2.8 s cold, **0.09 s** warm |
| `surface effect variants` test suite | 3m28s | **14 s** |

**What this means for the rest of the PR.** The precompile + disk layout cache + "Preparing surface shaders…" notice were built around a multi-minute cold start. At 11 s cold / 0.1 s warm they are no longer load-bearing and could be dropped to make this PR much smaller — say the word and I will strip them.

**And for macOS:** a 22k-line fragment shader with 861 copies of the stage chain is also a runtime cost (register pressure, instruction cache), so this may be a real part of the Apple Silicon slowness independently of the geometry stage. Worth measuring the mac numbers with *only* this commit before concluding the geometry stage was the culprit.

`SurfaceEffectVariantTest` now pins the GLSL size, so a reintroduced second return fails a test.

haraldsteinlechner:
## macOS numbers — the "what to check" from the PR description

Measured on Apple M1 (OpenGL 4.1), through the viewer's own scene graph
(`SurfaceEffectHarness.surfacesSg` → `createGroupedSgs`), on `24c87af8`. Pushed the harness
as `e86b9531` (`Tests.dll --bench`, docs in `docs/OPC-Surface-Effect-Benchmark.md`).

**#719's 13.9× reproduces on the viewer path**, so the "not measured on macOS" caveat can come off.

### HiRISE_VictoriaCrater, near-surface camera

`--eye 3376474.17 -324507.68 -121181.49`, fwd `-0.2756 -0.4158 -0.8667`. ms/frame, 9 batches per arm, 4× MSAA.

| resolution | lean | geometryStage | crossSection | both |
|---|---|---|---|---|
| 320×240   | 4.568 | 81.969 (**17.94×**) | 5.566 (1.22×) | 88.361 (19.34×) |
| 1024×768  | 6.106 | 83.882 (**13.74×**) | 7.608 (1.25×) | 92.645 (15.17×) |
| 1920×1200 | 6.920 | 87.787 (**12.69×**) | 8.990 (1.30×) | 95.846 (13.85×) |

### 1087_004779_MSLMST_0011 (the test fixture), bounding-box camera

| resolution | lean | geometryStage | crossSection | both |
|---|---|---|---|---|
| 320×240   | 0.257 | 0.993 (3.86×) | 0.399 (1.55×) | 1.150 (4.47×) |
| 1024×768  | 0.655 | 1.697 (2.59×) | 1.037 (1.58×) | 1.817 (2.77×) |
| 1920×1200 | 1.501 | 3.033 (2.02×) | 1.998 (1.33×) | 3.113 (2.07×) |

Zero correctness checks failed in either matrix.

### Findings worth carrying

- **The camera decides the answer.** Framing Victoria from its bounding box — whole dataset
  from above, coarse LoD — measures the geometry stage at **1.13×**. The same dataset from a
  near-surface camera measures **13.74×**. Any future benchmark has to pin a realistic camera
  or it will report a flattering number. The small fixture is too small to measure this stage
  with at all: on Victoria the delta is flat across a 30× pixel range (+77.4/+77.8/+80.9 ms,
  i.e. vertex-bound, matching #719), while on the fixture it scales with area.
- **`crossSectionClip` costs 1.22–1.30×**. Real and consistent, but
  ~60× smaller than the geometry stage. The discard axis earns its variant, but if the
  four-variant pool is ever revisited, this is the cheap axis, not the expensive one.
  (With `24c87af8` the startup argument against four variants is largely gone anyway:
  link 8.3 s cold for all four.)
- **`crossSectionClip` does not need a curtain texture** — `curtainEnabled` is independent
  of `clippingEnabled`.

### Two things that bit me, both worth fixing in this PR's area

1. **The GL-dependent Expecto tests deadlock on macOS rather than skipping.**
   `Render.context` creates `OpenGlApplication` on an Expecto worker thread; GLFW enters
   `-[NSApplication run]` and never returns. I lost 50 minutes to it before sampling the
   stack. `SurfaceEffectSwitchTest` is therefore not runnable on macOS as-is — the platform
   this PR exists to fix. The pushed commit adds `--gl-init-main`, which forces
   `Render.context` on the main thread before Expecto starts and makes them runnable.
2. **The cross-section polygon marks what is cut away, not what is kept.** `Surface.Sg`
   writes `if poly.Contains q2 then -d else d` and the shader discards on `< 0`. My first
   attempt used a polygon containing the terrain and discarded all of it — producing a fully
   black frame that still measured *faster than lean*. Worth a comment at the shader, and it
   is the reason every arm here asserts "same pixels as lean" before its timing is believed.

### Checked, and clean

The concern that dropping the geometry stage could starve `solarShadingLS` / `terrainSunShadow`
of their face normal (both are composed in *every* variant and read `localNormal`, which
`noFaceNormal` zeroes) **does not hold**. `SunLightEnabled` / `HasShadowMap` are bound off
`context.projectedImages`, which `wrapGisData` sets only when `body` *and* `projData` are
`Some` — and that `body` is the same `observedBody observationSystem` the gate uses. So
`SunLightEnabled ⟹ geometryStage`. The `noFaceNormal` comment is accurate; no change needed.

### Caveats

Wall clock only: `ITimeQuery` returns 0 on every batch on this driver (`RenderToken.TotalInstructions`
likewise; `DrawCallCount` works). One machine, one GPU, no Windows/Linux comparison, so
"geometry shaders are slow on Apple's GL-over-Metal" stays consistent with the data rather than
demonstrated by it. Arms are interleaved and warm-up discarded — measured naively, one `lean`
configuration drifted 4.383 → 0.979 ms/frame, larger than the effect being measured.

haraldsteinlechner:
## The monster is mostly gone

You were right: the start-up cost *was* the combinatorial explosion, not a missing cache. With the effect down to ~1000 lines, `d257ad35` removes most of what was built around it — **net -91 lines**.

**What went:**
- `precompile` (creating every variant's GL program upfront) and the `working`/`fallback` bookkeeping that guarded against a variant failing to link. That guard existed because a 22k-line shader really did fail to link once; at 830 lines it is not worth the code.
- the "Preparing surface shaders…" notice, and its row in the overlay.
- linking **all four** variants to build the shared input layout. It now comes from the **last variant alone** — the one that uses everything the others use. The others stay lazy: applying a layout does not link, and GL finds their programs on disk by effect id + layout hash. New test pins the assumption (no lean variant reads a uniform the full one does not).

**What stayed, and why it is not free to remove:** the layout cache. I measured it — the layout is exactly what GL keys its on-disk programs by, so without it *every* start pays the link again:

| | cold start | warm start |
|---|---|---|
| with layout cache (now) | 5.9 s link | **links nothing** |
| without it | 5.9 s | 5.9 s, every start |

It is ~45 mechanical lines (pickle two arrays, hash the effect id, read/write one 7 KB file). Happy to drop it too if you would rather have 6 s on every start than those lines — say so and it goes.

`data-surface-shaders="ready"` stays for the Playwright fixtures; it is now set after linking rather than after compiling.

All 5 switching tests and 6 variant tests pass, cold and warm.

haraldsteinlechner:
## Self-review before merge — two real defects found and fixed (`711de5e2`)

Ran two adversarial reviews over the diff (shader rewrites; the pool). The shader rewrites came back clean: all nine verified semantically identical, including the three traps (the `not(...)` inversion in `terrainSunShadow`, the shadowed `c` in `hoveredProjectionOutline` — correctly deleted rather than left in place — and `outcropTrace`'s inversion, where the 50-line body shows as pure context, so nothing moved into or out of the branch). The pool's per-surface gating was also verified complete: every reader of `LocalNormal` was enumerated against its gate, with no hole.

**Defect 1 — the rewrite missed the `LoDColor` that is actually in the stack.** It composes `OpcViewer.Base.Shader.LoDColor.LoDColor` (the NuGet package), not the copy I rewrote in the standalone demo viewer. I confirmed against the upstream source that the packaged one still has two returns — and it sits near the *start* of the stack, so it duplicated nearly everything after it. Switched to PRo3D's own copy, rewritten to one return:

| | before | after this commit |
|---|---|---|
| GLSL, lean | 830 lines | **623** |
| GLSL, full | 1016 lines | **805** |
| last stage inlined | — | **2×** (declaration + one use) — no duplication left |

**Defect 2 — the test guarding the new single-variant layout could not see the failure it guards.** Applying the shared layout is *lazy*, so a variant that does not fit throws on the **render thread** at first use (`[FShade] shader requests uniform X which is not part of layout`), mid-session. The subset assertions worked on `Effect.Inputs`/`Uniforms` — FShade's **pre-link** view — while the layout comes from the linked module, so they cannot prove the property. Replaced with a test that applies the real layout to all four variants and forces `Entries`. It passes, so the design is now demonstrated rather than assumed.

Also fixed: both assembly versions are back in the layout cache key (`Effect.link` reads more of the runtime than the effect id covers, so an Aardvark upgrade could load a mismatched layout); the cache file is written via a temp file so racing instances cannot leave a truncated one; `Log.stop` under `finally`; `ready` flipped outside the lock; the unused GL `ProgramExtensions` open dropped; and two docs that still claimed a surface with projection data "keeps the geometry stage from the start" corrected — it does not, it switches up.

**Known, accepted:** there is no longer a fallback if a variant fails to compile at runtime (the old `precompile` caught that once and pinned a working variant). The new layout test covers the failure mode that made it likely; a driver-side refusal would now surface as a render error rather than a silent degrade. Worth revisiting if macOS shows anything.

7 variant tests and 5 switching tests pass; cold link 5.3 s, warm links nothing.

PR #781, opened 2026-09-18
docs(gis): multi-image projection teaser, and three corrected claims

haraldsteinlechner:
## Why

The Multi-Image Projection page led with a stack screenshot, and three of its claims did not hold. This replaces the hero with a teaser generated from the shipped test data and fixes what was wrong.

teaser

## The teaser

`tests-ui/src/probe-teaser.ts` builds the four panels from `PRo3D.Resources.TestData`, so anyone can regenerate them:

| panel | measured |
|---|---|
| 1 DRACO mosaic only | **72.2%** of the visible body unobserved (black) |
| 2 plus 4 AFC images | **7.3%** |
| 3 second direction | 0.7% |
| 4 projector frustum | camera pulled back to 795 m |

Panels 1 and 2 are the same camera on the same body — that contrast is the whole point of the feature.

Cameras come from the scene's own **bookmarks**, written verbatim into `cameraView`. A bookmark stores all five view vectors, so this reproduces a chosen viewpoint exactly rather than re-aiming at the origin, which silently reframes the shot at close range.

## Corrections

Each checked against the code, not against the previous text:

- **Hovering does not preview the image in 3D.** `getSelectedImage` reads only `selectedImage`; hover feeds `hoveredProjection`, which draws the frustum wireframe and the footprint outline and nothing else. The old text described a workflow that does not exist.
- **"~10% of the body's pixels" had the wrong denominator** — `diffPng` divides by the whole frame (`image.ts:319`).
- **The fly-to control is a location-arrow icon**, not a paper plane.

The committed `multiProjection-hover.png` was shot on an **unbound scene**: its HUD reads `None xyz` / `conversion failed (set planet)`, which is the broken setup section 1 of the same page tells you to fix, and it was false-coloured with the Transfer Function on. Regenerated on a bound scene.

Section 3 now also states when the surface outline appears at all: the AFC-1 frames were taken from 6.7–8.4 km, which at that instrument's 5.53° field is a 650–810 m footprint over a 154 m body — it covers Dimorphos entirely, so no edge falls on it and no outline is drawn.

## ai/TESTING.md

Records what cost the most time, because it was not a rendering bug:

- **Never derive the background colour from the frame.** An unobserved mosaic region is pure black (0), the viewer clears to `#222222` (34). Sampling a corner breaks once the black body covers that corner; taking the mode breaks when black *is* the most common value. Both then report the body as absent — which is how a perfectly good frame was read as "`bodyPixels=9501`, essentially empty" and sent a session hunting a bug that did not exist. The real fault was a camera at 136 m from a body that needs ~199 m to fit the fov.
- **Check the subtended angle before blaming the harness:** `2·asin(r/d)`, against the fov `focal` implies.
- **aardvark.media hover** needs a real pointer crossing, and the row must be resolved by unique name — not by "nearest ancestor with a border", which is shared by every row.

`tests-ui/src/probe-bookmarks.ts` renders every bookmark of a scene and prints distance, subtended angle and coverage.

## Testing

Docs and test-harness only; no product code. Both probes run green against the local Dimorphos data and regenerate every image the page references.


haraldsteinlechner:
Closing as superseded. A full review of this branch found defects that make it not mergeable as-is:

**Docs — four false statements**, verified against the source:
- `MultiImageProjection.md` §3 claimed hovering does *not* project the image. It does: `ProjectedImageList-Model.fs:176-182` (`effectiveStack`) puts the hovered image on top of the stack as a preview. This PR replaced a correct sentence with an incorrect one — the single-image path I traced (`getSelectedImage`) is legacy and not what the viewer composes.
- `TimeAndSun.md` cited `--spice`; the flag is `--defaultSpiceKernel` (`CommandLine.fs:36`).
- `TimeAndSun.md` instructs the reader to click *Load Data*, which does not exist in the scenario it describes: `GisApp.fs:1267` populates the mission-time rows for a fresh viewer; the button appears only after loading a saved scene (`GisApp-Model.fs:195`).
- The hover figure caption ("the camera is 795 m out") describes a different screenshot than the one above it (198.8 m).

Plus: `MbiBased` documented as "MBI", "Import OPCs" as "Import OPC", the Reference System dropdown called "the planet", a contradiction introduced into `GisView.md`, a "one rotation" figure that is 5/6 of a turn, and three different rotation periods now stated across the repo.

**Harness — escalation residue.** The `probe-docs-shots.ts` rewrite carried in here contains ~80 lines that launch the app, scrape the camera out of HUD text, tear it down and relaunch with a hand-built view matrix (superseded by the bookmark approach in the same PR), ~60 lines of ArcBall-drag orbiting producing images referenced nowhere, a `hoverRow` this PR's own `ai/TESTING.md` documents as non-functional, and a HUD number parser broken in both locales. Four separately-tuned copies of `settled()`, none of which fail on timeout — they screenshot anyway, which is how a bad frame reaches `docs/images/`. `probe-bookmarks.ts` reproduces the corner-sampled-background bug that the same PR documents as the root cause of the original problem. And the teaser is not reproducible from shipped test data: it needs a private scene for its bookmarks.

Replaced by a branch rebuilt on the current `releases/6.3.0` (this one was based on 6.3.0-prerelease002, since superseded by the 6.3.0 release), keeping the figures and measurements that verified and dropping the rest.

